PR #8, opened 2016-01-09
docs: Add hubot::docker option for answers.yaml

logikal:
If you're trying to use answers.yaml to automate installation of stackstorm including chatops setup, `hubot::docker: True` is required, if you don't want to use the webgui installer. This adds documentation for that to the docs.


manasdk:
Thanks!

Will merge in a bit and use PR for testing

PR #11, opened 2016-01-11
Auto generate a list with overridable runner parameter attributes

Kami:
Instead of hard-coding it, we now auto generate it.

### DIFF
--- a/docs/source/_includes/runner_parameters_overridable_attributes.rst
+++ b/docs/source/_includes/runner_parameters_overridable_attributes.rst
@@ -1,0 +1,5 @@
+* default
+* description
+* enum
+* immutable
+* required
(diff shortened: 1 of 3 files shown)


Kami:
Tests fails because of two reasons:

1. Script looks for a matching branch in st2 (I mentioned this issue on Slack). I don't know if this is desired behavior or not. It doesn't seem so, at least for non-release branches, since it won't work if there is no corresponding branch in st2.

2. Current approach of retrieving branch name won't work since Travis performs a shallow clone. I had the same issue while working on pack tests, I literally spent hours trying to find an approach which works in all the scenarios. See here (https://github.com/StackStorm/st2contrib/blob/master/scripts/travis.sh#L3, https://github.com/StackStorm/st2contrib/blob/master/scripts/travis.sh#L18) for the approach I ended up using.

manasdk:
> Tests fails because of two reasons

Merged test failure fix on master and merged with master. Should be good to merge now since builds on master are passing https://travis-ci.org/StackStorm/st2docs/builds/101682092

Kami:
Merged, thanks.

PR #18, opened 2016-01-13
RFR: Add examples for ssl_cert and ssl_key specification in answers.yaml

lakshmi-kannan:
Thanks to @logikal for finding the solution and testing it. 

https://stackstorm-community.slack.com/archives/community/p1452212764011080

https://www.evernote.com/l/ANGHKRG1inlCw7HTRxikIRNOlYqVZ8pa79IB/image.png


manasdk:
+1

manasdk:
Maybe we should also CP the change to v1.2

Kami:
Nice :+1: 

Another option would also be to use line break characters (`\n`) or `|` YAML syntax.

It's also worth noting that the same "issue" exists for SSH key, but in that case we use line break characters.

PR #21, opened 2016-01-19
Add info about installing packs from private repo

lakshmi-kannan:


docs/source/packs.rst
```diff
@@ -51,6 +51,12 @@ By default packs are installed from the |st2| community repo. Use ``repo_url`` p
 
     st2 run packs.install register=all repo_url=https://github.com/StackStorm/st2incubator.git
 
+If you are using a private repo to host your pack(s),
+``repo_url=https://username:password@github.com/username/repository.git`` should get you going.
```
manasdk: Is this github specific? I don't see it in https://www.kernel.org/pub/software/scm/git/docs/git-clone.html#URLS

Even if it is github specific it should be fine to include so long as we call it out.

lakshmi-kannan: Yep, github specific. Let me add that to the explanation.

lakshmi-kannan: Fixed in bfd4c28

manasdk:
+1

PR #30, opened 2016-01-26
WIP: New installation docs

dzimine:
This PR adjusts the docs for upcoming new packages -  STORM-1979
I will use this doc to express "desired" state of installation; so please review and call out misalignments and mis-understandings.

- [x] Restructure Install section
- [x] Describe "big picture" - single node deployment "end-result"
- [x] Guide structure for both Ubuntu and RHEL/CentOS
        * a single guide for all supported OS, [Sensu style](https://sensuapp.org/docs/latest/install-sensu)
- [x] Migrate Installation Guide steps from st2-packages for Ubuntu
- [x] Add RHEL instructions
- [x] Update "other" installation options (puppet, chef, ansible, salt, docker)
- [ ] Updates and Upgrades (with special note on how to migrate from any version)


dzimine:
image

here's the structure, quick feedback welcome @lakshmi-kannan

dzimine:
### After taking @lakshmi-kannan feedback:

image

dzimine:
@lakshmi-kannan @dennybaa @armab Let's review the reference deployment and structure of installation guide.

docs/source/install/overview.rst
```diff
@@ -0,0 +1,62 @@
...
+
+    StackStorm single-box reference deployment.
+
+.. source https://docs.google.com/drawings/d/1X6u8BB9bnWkW8C81ERBvjIKRfo9mDos4XEKeDv6YiF0/edit
+
+.. todo:: keep image live `on google.drive <https://docs.google.com/drawings/d/1X6u8BB9bnWkW8C81ERBvjIKRfo9mDos4XEKeDv6YiF0/edit>`_ while WIP but copy a final .png under _static
+
+1. st2* services
+----------------
+st2* services provide main stackstorm functionality. They are located at ``/opt/stackstorm/st2``, share a dedicated Python virtualenv, and configured by /etc/st2/st2.conf.
```
enykeev: Are you positive about their location? Because I'm pretty sure they ain't there on our new packages (https://github.com/StackStorm/st2-packages/blob/master/packages/st2bundle/rpm/st2api.service#L11).

docs/source/install/overview.rst
```diff
@@ -0,0 +1,62 @@
+Overview: Single-box Reference Deployment
+=========================================
+
+.. todo:: short intro paragraph: this is reference deployment for single box, follow installation
+    guide to get here.
+
+.. figure:: https://docs.google.com/drawings/d/1X6u8BB9bnWkW8C81ERBvjIKRfo9mDos4XEKeDv6YiF0/pub?w=960&amp;h=720
```
enykeev: image 2016-01-29 at 12 05 23 pm

docs/source/install/overview.rst
```diff
@@ -0,0 +1,62 @@
...
+    * st2notifier service supports :doc:`/chatops/notifications`. `TODO: specify or refer what needs configuration to enable it`.
+    * st2garbagecollector** is an optional service to periodically delete old execution history data from the database, per settings in st2.conf.
+    * st2auth is an authentication service with the REST endpoint. A variety of auth backends is available; see :doc:`/authentication`. Reference deployment uses <flat file backend `https://github.com/StackStorm/st2-auth-backend-flat-file`>_ backend.
+    * st2api is REST API web service endpoint; it also serves webhooks for webhook triggers.
+
+
+2. st2-mistral
+--------------
+Mistral is a workflow service component that StackStorm uses for long-running workflows. It is installed under ``/opt/stackstorm/mistral``, runs in a dedicated Python virtualenv, and configured by /etc/mistral/mistral.conf. ``mistral`` reaches out to st2api for action execution requests. ``mistral-api`` is an internal end-point accessed by st2actionrunner and st2notifier. In a single-box deployment it is restricted to localhost.
```
enykeev: Not sure it's where you think it is either (https://github.com/StackStorm/st2-packages/blob/master/packages/st2mistral/debian/postinst#L7)

docs/source/install/overview.rst
```diff
@@ -0,0 +1,62 @@
...
+3. st2client
+-------------
+st2client is the `st2` CLI and python bindings for StackStorm API. To configure CLI to point to the right API, authenticate options, suppressing unsecure warnings for self-signed certificates and othe cool stuff see :doc:`/cli`.
+
+
+4. ngnix for WebUI and SSL termination
+--------------------------------------
+* **nginx** provides SSL termination, redirects HTTP to HTTPS, serves WebUI as static HTML, and reverse-proxies REST API endpoints to st2* web services.
+
+* **StackStorm WebUI** (st2web and flow) are installed at ``/opt/statckstorm/static/webui`` and configured by `webui/confi.js`. They are are HTML5 applications, served as static HTML and calling StackStorm st2auth and st2api REST API endpoints. With ngnix proxying st2auth and st2api through 443 HTTPS port it's all good. Overwise proper CORS configuration is required both on st2 and `webui/config.js`. TODO:XXXX polish text here.
```
enykeev: > `webui/confi.js`

typo

> With ngnix proxying st2auth and st2api through 443 HTTPS port it's all good.

The idea is that static assets should share the same virtual host (precisely, a combination of host and port) with API reverse proxy. And when they're not, both party should be aware of that (viz. st2web should point to proper locations and this locations should respond with CORS headers mentioning host st2web deployed at).

docs/source/install/overview.rst
emedvedev: Am I the only one who thinks this pic makes St2 look overly complicated?

enykeev: Well, I think that ST2 **is** overly complicated. With that in mind, yes, the picture communicates this impression pretty well.

emedvedev: I just don't think we should make this first impression on potential users. It's a good pic when you need to explain deployment to power users or enterprise customers who know their stuff well and played with St2 for a while, but it's arguably not very good for newcomers, so maybe it shouldn't be here. @dzimine: thoughts?

manasdk: @enykeev - once you make such a definitive claim you also gotta suggest an alternative :)

Looking at the picture I wonder if can can start pulling boxes together. If anyone cares to discuss I would love to chat offline and see where we can simplify

enykeev: No @manasdk, I do understand why we end up where we end up. There was a looong chain of decisions and each of them we're more or less valid. Dunno, may be "overly" is no the right word... "sufficiently complicated". Anyway, the point is, st2 is huge and we should not hide this fact from our users.

emedvedev: Nobody wants to discuss architecture here and nobody says we should hide anything from users. :) It's just that St2 looks a bit messy in this pic, that's all I'm saying, and I wonder if we can make the pic better to turn it into a better first impression. Making St2 better goes into other discussions. :)

docs/source/install/overview.rst
enykeev: Disregard, found a note of you guys discussing it.

armab: @enykeev is right, any services, source code, python helpers never were in `/opt/stackstorm`.
* `/usr/lib/python2.7/dist-packages/...` - old packages.
* `/usr/share/python/st2/` - new packages.

As I understand, placing everything, including source code, upstart scripts in `/opt/stackstorm` is not really the right way from the packaging point of view.
@dennybaa should know "the right way" better, please comment.

docs/source/install/overview.rst
```diff
@@ -0,0 +1,62 @@
...
+
+.. source https://docs.google.com/drawings/d/1X6u8BB9bnWkW8C81ERBvjIKRfo9mDos4XEKeDv6YiF0/edit
+
+.. todo:: keep image live `on google.drive <https://docs.google.com/drawings/d/1X6u8BB9bnWkW8C81ERBvjIKRfo9mDos4XEKeDv6YiF0/edit>`_ while WIP but copy a final .png under _static
+
+1. st2* services
+----------------
+st2* services provide main stackstorm functionality. They are located at ``/opt/stackstorm/st2``, share a dedicated Python virtualenv, and configured by /etc/st2/st2.conf.
+
+    * st2sensorcontainer runs sensor from ``/opt/stackstorm/packs``.
```
armab: As I understand from new path hierarchy, it should be `/opt/stackstorm/st2/packs`

armab:
> `Update "other" installation options (puppet, chef, ansible, salt, docker)`

^^ Is not possible until we ship all packages in production and polish/stabilize the things.

At least for `Ansible` I know that I should completely wipe entire https://github.com/stackstorm/ansible-st2 and rework it with new packages. Polish/test/release.
So configuration management will be absolutely different story. But later, after the packages.

docs/source/install/overview.rst
dzimine: Thanks for the comments. I'll edit the pic but KEEP it as complex as it is. This is not a "first" picture folks see; the simpler one is in the overview. Somewhere ( and I assume, here) we need a good wiring diagram for someone who wants to understand how things are really related.

docs/source/install/overview.rst
dzimine: That's how it should be. Discussed, decided. Thanks for noticing.

Special note on Mistral: we ship StackStorm's version of Mistral, we stand behind it, and we are a vendor who is delivering it. The fact that there's an upstream, and we have fork, and keep it close to upstream is irrelevant under-the-hood details. Please treat it like this.

dzimine:
>  ^^ Is not possible until we ship all packages in production and polish/stabilize the things.
@armab than we ditch this: https://docs.stackstorm.com/install/ansible.html and https://docs.stackstorm.com/install/puppet.html and other config from the doc till they work. On the other hand, if it continues to work, but not with the new packages, it can stay. 

Up to the discussion.

docs/source/install/guide.rst
```diff
@@ -0,0 +1,225 @@
...
+  .. code-block:: bash
+
+    # Create an SSH system user
+    useradd stanley
+    mkdir -p /home/stanley/.ssh
+    chmod 0700 /home/stanley/.ssh
+
+    # Generate ssh keys on StackStorm box and copy over public key into remote box.
+    ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
+    cp ${KEY_LOCATION}/stanley_rsa.pub /home/stanley/.ssh/stanley_rsa.pub
```
lakshmi-kannan: Delete this line. ssh-keygen already places the public key in right place.

```
root@pkgtestingbox:/home/vagrant# ssh-keygen -f /home/stanley/.ssh/stanley_rsa -P ""
Generating public/private rsa key pair.
Your identification has been saved in /home/stanley/.ssh/stanley_rsa.
Your public key has been saved in /home/stanley/.ssh/stanley_rsa.pub.
```

docs/source/install/guide.rst
```diff
@@ -0,0 +1,225 @@
...
+  .. code-block:: bash
+
+    echo "stanley    ALL=(ALL)       NOPASSWD: SETENV: ALL" >> /etc/sudoers.d/st2
+
+* Adjust configuration in ``/etc/st2/st2.conf``:
+
+  .. sourcecode:: ini
+
+    [system_user]
+    user = stanley
```
lakshmi-kannan: default st2.conf ships with right configuration. remove these lines.

docs/source/install/guide.rst
```diff
@@ -0,0 +1,225 @@
...
+    ssh_key_file = /home/stanley/.ssh/stanley_rsa
+
+Check it all works
+~~~~~~~~~~~~~~~~~~
+
+  .. code-block:: bash
+
+    st2 --version
```
lakshmi-kannan: By the time we are here, we are root if I followed the guide strictly. Let's maybe fix the SSH keypair commands to use sudo on every line instead of root or add a line at end to logout as root.

docs/source/install/guide.rst
```diff
@@ -21,31 +18,62 @@ Install Dependencies
...
+
   .. code-block:: bash
 
-    sudo apt-get update
+    sudo apt get update
     sudo apt-get install -y mongodb-server rabbitmq-server postgresql
 
+  .. rst-class:: centos
+
+    CentOS/RHEL
```
armab: Looks like the change is to include both rpm/deb instructions in one big guide.

I worked on testing new packages under `RHEL7/CentOS7` and draft with minimal installation instructions is: https://gist.github.com/armab/2ab4d2ccc592db42243

Some points to avoid this merge if possible: 
- There are already slightly different instructions in logic for `rpm`, comparing to `deb`. Like for `el7` you need additional commands, which are not required for `ubuntu`.
- There is even a chance that instructions for `el6` comparing to `el7` will slightly differ. I didn't tested `el6`, focusing on `el7` first, but I suspect where could be differences. That means - more "branching" in one doc, where you need to cover more special cases.
- A lot of text makes instructions always looking harder both in complexity and readability/usability.

Yes, it would be less easy for us to manage and keep in sync repeating/duplicated pieces, but I believe it's overall easier for user with separated docs, keeping in mind that the goal is successful and more painless/easy installation.

docs/source/install/guide.rst
```diff
@@ -21,31 +18,62 @@ Install Dependencies
...
 Install MongoDB, RabbitMQ, and PostgreSQL.
 
+  .. rst-class:: ubuntu
+
+    Ubuntu/Debian
+
   .. code-block:: bash
 
-    sudo apt-get update
+    sudo apt get update
```
armab: `-` is missing here because of some typo.

docs/source/install/deb.rst
```diff
@@ -30,13 +31,12 @@ Install MongoDB, RabbitMQ, and PostgreSQL.
 Setup repositories
 ~~~~~~~~~~~~~~~~~~~
 
-  .. code-block:: bash
+The following script will detect your platform and architecture and setup the repo accordingly. It'll also install the GPG key for repo signing.
+Currently we support ``Ubuntu Trusty``, ``Debian Wheezy`` and ``Debian Jessie``.
```
lakshmi-kannan: Don't complain Debian *.

docs/source/install/deb.rst
```diff
@@ -80,20 +80,21 @@ For remote linux actions, SSH is used. It is advised to configure identity file
...
-    cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys
-    chmod 0600 /home/stanley/.ssh/authorized_keys
-    chown -R stanley:stanley /home/stanley
+    sudo sh -c 'cat /home/stanley/.ssh/stanley_rsa.pub >> /home/stanley/.ssh/authorized_keys'
+    sudo chmod 0600 /home/stanley/.ssh/authorized_keys
+    sudo chown -R stanley:stanley /home/stanley
 
     # Enable passwordless sudo
-    echo "stanley    ALL=(ALL)       NOPASSWD: SETENV: ALL" >> /etc/sudoers.d/st2
+    sudo sh -c 'echo "stanley    ALL=(ALL)       NOPASSWD: SETENV: ALL" >> /etc/sudoers.d/st2'
```
lakshmi-kannan: @dennybaa has some packaging changes that does this as part of packaging.

dzimine: I am not sure if it's good to make it part of the package. Even if we do, this needs to stay here to highlight the needed steps of setting up st2 system user.

docs/source/install/deb.rst
dzimine: copy-paste from v1.3. Ok, removing.

dzimine:
I am merging. Let's finish together on master.

docs/source/install/deb.rst
dennybaa: @dzimine it should be documented but not suggested as an action.
please review https://github.com/StackStorm/st2-packages/pull/195 . Since @enykeev @armab and me I +100 for this. We have to permissions story sorted out and documented, no we have none of that except those vague steps which are clear only for us.

1. actionrunner should operate as `system_user`, to avoid excessive privilege elevation https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/system/action.py#L79-L93 . Everytime when we are `!= system_user` we use sudo, that doesn't make sense.
2. we need permission allowance cross st2api and st2actionrunner to enable flow to save packs.

The current not documented approach how to run and how privileges are elevated is only a mess and a trouble.

PR #42, opened 2016-02-04
Use sudo instead of sudo su

DoriftoShoes:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454628411005118

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454628530005124

dzimine:
ok, taking this, will do other @lakshmi-kannan suggestions on top.

PR #43, opened 2016-02-05
Sprinkle some -y's on the apt-gets

DoriftoShoes:
This is so the automated script matches the documentation.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454632919005142

dzimine:
I'd rather diverge from script here. Let the users stop and see what's happening. They can type -y.

PR #50, opened 2016-02-05
Chatops instructions

emedvedev:
Instructions for Hubot installation on a clean machine. Please confirm it works; it did on my testing machine but it’s all kinds of messed up. I’ll do some more testing, too.

Honestly, compared to puppet+st2installer this is a huge UX regression and takes ~15 minutes to set up assuming everything runs smooth (and for new users it almost never does). We should think about making a bootstrap script from the very beginning, probably.


dzimine:
@emedvedev thank you; this is good, but doesn't fit here. The tone of the guide is instructional and scriptable; and it applies to one particular reference installation. Chat providers are still a variance. 

I'd like to save this section for placing somewhere in the reference; but rework it for here so that one can follow the steps exactly to get chatops running.

dzimine:
> Honestly, compared to puppet+st2installer this is a huge UX regression and takes ~15 minutes to set up assuming everything runs smooth (and for new users it almost never does).
Yeah. Took me 1/2 a day to set up the ChatOps. Ok, + write scripts + write instructions, but the point stays. A package will help.

emedvedev:
Hey, a pretty great update and a solid script! I like it.

PR #58, opened 2016-02-10
Add a toctree to development/index.rst with links to other pages.

jon-middleton:
I've found the code_structure, pack_testing and testing hard to find
as they were hidden as small links within the body of text. So this PR
adds a toctree with the 3 other pages under development.


dzimine:
Thanks Jon, agreed, merging.

PR #59, opened 2016-02-10
Add new page covering pack installation via ChatOPS

jjm:
Also cover the configuration required to enable auto-deployment.

This covers the additions that were merged with PR#2465 in st2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ st2
 virtualenv
 build
 *.py[cod]
+
+# Editor Saves
+*~
+\#*\#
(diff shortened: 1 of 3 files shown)


dzimine:
Thank you for the contribution! off it goes. 
Could you please open another PR against v1.3 so that it appears on the current / 1.3 docs, not just "latest"?

jjm:
Sure, just to confirm tho. Is PR#2465 going to be included in 1.3.x? I thought it would get released with 1.4.

dzimine:
Oh, you're right, my bad. PR#2465 is 1.4. So never mind.

PR #61, opened 2016-02-12
update doc parameters for custom SSL cert/key

codec:
This updates the docs/FAQ to reflect the current parameters to use custom SSL certificate and key.


emedvedev:
We should probably backport it, too.

PR #64, opened 2016-02-18
Update AWS install instructions

joshuajabbour:
The Marketplace UI seems to have changed since these were written. These still aren't perfect, mainly because the supported AMIs are no longer available (that I can find). But at least now the steps work with the default Ubuntu 14.04 base AMI.

### DIFF
--- a/docs/source/install/all_in_one.rst
+++ b/docs/source/install/all_in_one.rst
@@ -105,29 +105,22 @@ Visit the setup URL output on your command line by entering the address in your 
 Amazon Web Services (AWS)
 =========================
 
-We have currently performed testing and certified that StackStorm Installer works properly on the following AMI Images:
-
-* CentOS 6     - `ami-1255b321`
-* CentOS 7     - `ami-d440a6e7`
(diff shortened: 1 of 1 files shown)


dzimine:
Thank you for the PR! If you could also: 
* update the AMI for the default ubuntu `ami-9abea4fb` - we have switched our testing to it and know it work.

dzimine:
@joshuajabbour Thank you for the contribution! It's tough to stay current without community help. Really appreciate.

joshuajabbour:
Well, actually if you start from [here](https://cloud-images.ubuntu.com/releases/14.04/release-20160114.5/), the instructions seem to be much more accurate. That however is not the AWS Marketplace. I'll take another pass at this when I get a chance...

joshuajabbour:
This probably shouldn't have been merged. As I mentioned above, the actual steps in the current instructions do appear to be accurate. There's just not enough detail on how to get started...

dzimine:
I should pay more attention. Will handle it later tonight

PR #65, opened 2016-02-19
Docs explaining webhook normalization

manasdk:


manasdk:
Travis failure unrelated and fixed on st2 already.

PR #69, opened 2016-02-19
rules.rst: Fix sub bullet points in Rule Structure section

gaelL:
See https://docs.stackstorm.com/rules.html#rule-structure
Sub bullet point are displayed inline.

Tried this syntax http://docutils.sourceforge.net/docs/user/rst/quickstart.html#id18

Seems display well on my github branch https://github.com/gaelL/st2docs/blob/rules/docs/source/rules.rst


Kami:
Good catch. Thanks.

emedvedev:
Again, thanks a lot :)

PR #70, opened 2016-02-22
Document operator changes

Kami:
Documentation changes for https://github.com/StackStorm/st2/pull/2519.


Kami:
Just FYI - using tables for those descriptions wasn't such a good idea since the description field doesn't wrap and it's hard to read.

We should figure out how to make it wrap (maybe CSS hack) or switch back to non-table display.

PR #72, opened 2016-02-24
Fix links to all-in-one for new installation methods

enykeev:


enykeev:
Full disclosure, I haven't tested it locally. I was lazy to spin up a vm and had an issue building inotify on osx I didn't want to deal with.

```Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/0d/517h6xjd1790vyf1wrdl9xch0000gn/T/pip-build-IonqH8/pyinotify```

dzimine:
+1 looks right, better than it was before :)

PR #74, opened 2016-02-25
correct stanley sudoers.d permission

dennybaa:
This is a MUST. Otherwise privileged access is broken if there's no direct ability to login as `root`. Since after breaking sudo, there's no chance for `sudo -i` or `sudo su` anymore. It's haters issue)


lakshmi-kannan:
This also requires a corresponding change in bootstrap scripts- https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh and https://github.com/StackStorm/st2-packages/pull/181.

lakshmi-kannan:
Picked up change in https://github.com/StackStorm/st2docs/pull/75. Closing.

PR #75, opened 2016-02-28
RFR: Install docs for RHEL 6/7

lakshmi-kannan:
## Disclaimer

If you are not a stormer trying out these instructions, DO NOT expect support on community. 

## TODO

* [ ] List boxes used to test these instructions in docs. Salvage from https://github.com/lakshmi-kannan/st2-test-ground/blob/master/Vagrantfile.
* [x] Hand test these instructions manually on a box cc: @dzimine 
    1. [X] RHEL7
    2. [ ] RHEL6 - Packages don't work on RHEL6 because of missing libffi-devel
* [x] Chatops section requires rules to be registered but st2ctl reload doesn't do it by default. See 92a6afa


docs/source/install/rhel7.rst
```diff
@@ -0,0 +1,387 @@
...
+Install Dependencies
+~~~~~~~~~~~~~~~~~~~~
+
+Install MongoDB, RabbitMQ, and PostgreSQL.
+
+  .. code-block:: bash
+
+    sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+    sudo yum -y install mongodb-server rabbitmq-server
+    sudo systemctl start mongod rabbitmq-server
```
dzimine: 1) systemctl vs service - for the commands . I checked that most folks out there do so, even those who do separate docs for each flavor, e.g. docker, https://docs.docker.com/v1.8/installation/rhel/

2) Let's add instructions on making service restart after computer reboot - Anthony reported as a bug in new instructions as it caused him grief and support call to figure why st2 is not up (turned out mongo didn't come up after reboot).
`sudo chkconfig rabbitmq-server on`

docs/source/install/rhel7.rst
```diff
@@ -0,0 +1,387 @@
...
+  To control Windows boxes, configure access for :doc:`Windows runners </config/windows_runners>`.
+
+* Adjust configuration in ``/etc/st2/st2.conf`` if you are using a different user or path to the key:
+
+  .. sourcecode:: ini
+
+    [system_user]
+    user = stanley
+    ssh_key_file = /home/stanley/.ssh/stanley_rsa
+
```
dzimine: Based on @emedvedev feedback - "if I follow literally at this point I am root", and instructions down below don't assume it... what about adding `sudo su` at the beginning of this block and `exit` here? Not 100% certain.

docs/source/install/rhel7.rst
```diff
@@ -0,0 +1,387 @@
...
+Start Services
+~~~~~~~~~~~~~~
+* Start services ::
+
+    sudo st2ctl start
+    sudo systemctl start mistral
```
dzimine: This is not needed. st2ctl starts mistral just fine.

docs/source/install/rhel7.rst
lakshmi-kannan: 2) Yep. I hit that myself. We need two changes - one in docs and another in script. This is tedious. More like a general point. 

1) I don't get you. Are you saying s/systemctl/service? I noticed that in docker instructions myself. I can make that change.

docs/source/install/rhel7.rst
lakshmi-kannan: This is what I said too :P. You decided not to do sudo in front of each command. Script actually uses sudo in front of each command.

docs/source/install/rhel7.rst
lakshmi-kannan: Good point, fix incoming.

docs/source/install/rhel7.rst
dzimine: Ok, let's do use sudo in front of each for full consistency. I was just lazy to learn how to sudo `>>` redirects .

docs/source/install/rhel7.rst
```diff
@@ -0,0 +1,387 @@
...
+
+    # Copy and enable StackStorm's supplied config file
+    sudo cp /usr/share/doc/st2/conf/nginx/st2.conf /etc/nginx/conf.d/
+
+    # Disable default_server configuration in existing /etc/nginx/nginx.conf
+    sudo sed -i 's/default_server//g' /etc/nginx/nginx.conf
+
+    sudo systemctl restart nginx
+
+If you modify ports, or url paths in nginx configuration, make correspondent chagnes in st2web
```
dzimine: Problem on AWS Redhat 7, ami: RHEL-7.2_HVM_GA-20151112-x86_64-1-Hourly2-GP2 (ami-775e4f16)

### Symptoms: 
After restarting nginx web ui login comes up but the https://<host>/api and https://<host>/auth ports return 500 bad gateway. Checking /var/log/nginx/ssl-st2webui.error.log, there are errors: 
```
2016/02/27 22:44:59 [crit] 24354#0: *16 connect() to 127.0.0.1:9101 failed (13: Permission denied) while connecting to upstream, client: 69.181.118.28, server: , request: "GET /api/ HTTP/1.1", upstream: "http://127.0.0.1:9101/", host: "52.36.162.164"
2016/02/27 22:45:09 [crit] 24354#0: *16 connect() to 127.0.0.1:9100 failed (13: Permission denied) while connecting to upstream, client: 69.181.118.28, server: , request: "GET /auth/ HTTP/1.1", upstream: "http://127.0.0.1:9100/", host: "52.36.162.164"
```

Apparently, SELinux problem, common on Fedora/CentOS/RHEL (see https://stackoverflow.com/questions/23948527/13-permission-denied-while-connecting-to-upstreamnginx) 

`setsebool -P httpd_can_network_connect 1` fixed the problem. May be a more refined policy described in stackoverflow may be a more proper solution, but I'd say, up to a user now. 

It may be better to add "Troubleshooting" section and add references there, as we don't clearly know yet if it's a distro or image problem.

docs/source/install/rhel7.rst
```diff
@@ -0,0 +1,387 @@
...
+configuration at ``/opt/stackstorm/static/webui/config.js``.
+
+Use your browser to connect to ``https://${ST2_HOSTNAME}`` and login to the WebUI.
+
+Setup ChatOps
```
dzimine: if this is same for all OS which support docker, modular "how to install docker on your OS", we shall factor this block out of here. But can do it later.

docs/source/install/rhel7.rst
lakshmi-kannan: Script is already adjusting SELinux policies https://github.com/StackStorm/st2-packages/pull/188/files#diff-3277f0bb4e4f9313bbdb9f2246b0a7b3R8. I forgot to copy them over to docs. Will add a section on SELinux at the top.

docs/source/install/rhel7.rst
lakshmi-kannan: +1 for later.

docs/source/install/rhel7.rst
lakshmi-kannan: As per offline discussion, punted on systemctl -> service change.

docs/source/install/rhel7.rst
lakshmi-kannan: Fixed in 5473dc21e346030a3cef9fcd26b4af1217ee504d

dzimine:
Ok, ran st2-self-check, with following workarounds: 

1) hack path in the script st2-self-check and placed it there https://gist.github.com/dzimine/61a8c055c5e2917d581b
2) st2-setup-examples and st2-setup-tests scripts need to be packaged up to `/opt/stackstorm/st2/bin`, 
3) virtualenv is not on RHEL7 box by default thus st2-setup-tests got crippled (not failed, bitch!), so be sure it is in place prior to running it... or check in script.

Two failures, but both in tests themselves: 
1) tests.test_packs_pack - failed on  cffi>=1.4.1 because there's no `gcc` around
2) tests.test_timer_rule - apparently `Rule "fixtures.tests.interval_timer" is not found.` but looks like rule have not been loaded - 
https://github.com/StackStorm/st2/blob/master/tools/st2-setup-tests#L50-L54 
how it ever worked...

dzimine:
now, :+1:

docs/source/install/rhel7.rst
lakshmi-kannan: Fixed scripts in e66d0f6. Fixed docs in 0495bb5.

armab:
Seems the order for EL6/7 instructions here is different comparing to initial draft: https://gist.github.com/armab/2ab4d2ccc592db422439

From the [`notes`](https://gist.github.com/armab/2ab4d2ccc592db422439): 
> Because st2mistral requires more steps, st2 installation comes first to bring "fast win" feeling to user, when st2 could be installed in really 2 steps.


I really would suggest this order (`st2` -> `st2mistral` -> ...).
Bringing up `st2` *before* `st2mistral` is better/easier from user experience point of view, when they have something installed and working, before proceeding with further steps.

I guess that between goals we all keep in mind something like: Decrease user frustration, decrease "bounce rate" when users just leave during their first software evaluation experience.
Talking about psychological aspect, if user just made `st2` working early and something fails on step later, they less likely to opt-out (%), comparing to existing approach. 

Just opinion FYI to consider ^^ @dzimine and @lakshmi-kannan
(Also that requires changing order for `deb` instructions)

armab:
Also from the [notes](https://gist.github.com/armab/2ab4d2ccc592db422439):

> Added a two-words clarification what is StackStorm and what is Mistral. Expect users which "heard something" about IFTTT for Ops to visit installation page more frequently, comparing to other pages in doc.

I would suggest to include something like this in docs (2 words).

Just place yourself on user's side when you evaluate any software for the first time.
It's usual `I don't know what I'm doing` behavior. It may be better to bring them idea what they're installing and why, at a high level.


@lakshmi-kannan @dzimine 

Honestly, I didn't know what is mistral and why it's needed at all for pretty long time during the first days of `st2` evaluation. Even more, when I installed `st2` for some of my clients, they never needed mistral features (it's more advanced/optional thing for the regular user).
^^ Just FYI some facts from the reality behind this suggestion.

dzimine:
@armab thanks, I appreciate the comments re mistral, I am not dismissive, facts acknowledged, however:
> Added a two-words clarification what is StackStorm and what is Mistral.
mistral is not optional. Please bear with me on this for now, 
and I'll spend more time one day sharing facts from other users, and explaining the thinking behind it.
For now, give me smooth experience that ensures that mistral is always on and working.

dzimine:
and yes let's add few words on "what is behind", explanation of the pieces, etc. But in a separate commit.

docs/source/install/rhel6.rst
```diff
@@ -0,0 +1,287 @@
...
+
+* If previous command returns enforcing, then run the following commands to adjust SELinux policies:
+
+    .. code-block:: bash
+
+        # SELINUX management tools, not available for some minimal installations
+        sudo yum install -y policycoreutils-python
+
+        # Allow rabbitmq to use '25672' port, otherwise it will fail to start
+        sudo semanage port --list | grep -q 25672 || sudo semanage port -a -t amqp_port_t -p tcp 25672
```
dzimine: suggest we just tell them `sudo semanage port -a -t amqp_port_t -p tcp 25672`, save check for the script. It's cleaner what user is doing, and if doesn't work it'll say "already set".

docs/source/install/rhel6.rst
```diff
@@ -0,0 +1,291 @@
...
+Install MongoDB, RabbitMQ, and PostgreSQL.
+
+  .. code-block:: bash
+
+    sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-6.noarch.rpm
+    sudo yum -y install mongodb-server rabbitmq-server
+    sudo service mongod start
+    sudo service rabbitmq-server start
+    sudo chkconfig mongod on
+    sudo chkconfig rabbitmq-server on
```
dzimine: How about grouping  "making service stay on" under same comment
```
# Makes Mongo and RabbitMQ start on boot
sudo chkconfig mongod on
sudo chkconfig rabbitmq-server on
```

lakshmi-kannan: Sorry, all these changes can be made after merging. Let's focus on correctness of instructions first.

docs/source/install/rhel6.rst
lakshmi-kannan: I removed -q from the command which only suppresses the output and looks at exit code instead. See b786d195e8d87f0e877a7f414e7ead36c52eb6ea. Also, clarified some things about SELinux.

docs/source/install/rhel6.rst
```diff
@@ -0,0 +1,291 @@
...
+    -days 365 -nodes -subj "/C=US/ST=California/L=Palo Alto/O=StackStorm/OU=Information \
+    Technology/CN=$(hostname)"
+
+    # Copy and enable StackStorm's supplied config file
+    sudo cp /usr/share/doc/st2/conf/nginx/st2.conf /etc/nginx/conf.d/
+
+    # Disable default_server configuration in existing /etc/nginx/nginx.conf
+    sudo sed -i 's/default_server//g' /etc/nginx/conf.d/default.conf
+
+    sudo service nginx restart
```
dzimine: should be `nginx start` ? no point of restarting in context and extra "FAILED" on the screen.

lakshmi-kannan: restart doesn't fail. if it's not started, it starts.

PR #78, opened 2016-02-29
Add more prominent link to the YAQL evaluator website

Kami:
It turns out we already link to that website, but we only link to it on the Mistral page and there are no links to it in the YAQL section of our documentation.


dzimine:
+1

PR #80, opened 2016-02-29
Add instructions for installing chatops from deb

enykeev:


dennybaa:
:+1:

lakshmi-kannan:
We also need a corresponding change in bootstrap script for ubuntu. Otherwise, script users won't get chatops. I'll do testing with these instructions and report. This PR is for master. If you want 1.3 users to pick this up, change should be in 1.3 branch. Of course, we'll need 1.3 packages too.

dzimine:
I am merging it on master to sort it out there. Than we can cherrypick to 1.3 later this week.

PR #84, opened 2016-03-03
Point enterprise installation to stable repos

dzimine:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457043073005877

armab:
@dzimine @lakshmi-kannan 

I guess this PR is intentional (for future), but just a reminder to be sure: 
We don't have packages in `enterprise` stable yet, only `enterprise` staging.

This should be possible after CI promotion `staging` -> `production`.

dzimine:
I think @lakshmi-kannan or @DoriftoShoes moved them manually, if not, we should. 
This is driven by the fact that the same access key can not work for both `enterprise` and `enterprise-staging`. And because our current approach License==key, each user begins to get new License that opens two locks - old AIO installer download site, and only ONE of package cloud repos, we have to choose. And once chosen, it's not easy to re-issue the key. 
Sorry if the explanation is confusing, we can pick it on slack.

armab:
Yeah, `StackStorm/enterprise` repo is currently empty.

lakshmi-kannan:
@armab I didn't get to move the packages to prod. Can you do that?

armab:
@lakshmi-kannan Sure, I can upload them manually on Monday.

armab:
Seems @lakshmi-kannan already figured it out.
I just uploaded missed packages for `debian wheezy` and `debian jessie`, probably oversight.

PR #93, opened 2016-03-07
ssh owning inconsistencies

dennybaa:
`authorized_keys` is a list of **public** keys which are allowed to use key-based logins. Setting mode to `0600` doesn't make any sense, even if a user can not read `/home/stanley/.ssh/authorized_keys`, well he can always read `/home/stanley/.ssh/stanley_rsa.pub` which has been created without such restrictive permissions. So setting `0600` for `authorized_keys` is quite ambiguous...


lakshmi-kannan:
good catch. please make corresponding bootstrap script changes when you make such changes. https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el6.sh#L85

lakshmi-kannan:
Needs to be merged to v1.3 as well.

PR #94, opened 2016-03-09
Install chatops package

manasdk:
Done on ubuntu but not on rhel.

Unsure if this step is needed but it would seem that this is correct.


manasdk:
Based on https://github.com/StackStorm/st2-packages/blob/470a32582e64cbdc5222f3b7b0271d38fae775b5/scripts/st2bootstrap-el6.sh#L201 it would seem I am correct.

dzimine:
Oh... I can't believe we missed it. Thank you!

dzimine:
Can you pls cherrypick to v1.3, too?

manasdk:
cherrypick https://github.com/StackStorm/st2docs/pull/95

PR #105, opened 2016-03-23
Update aliases.rst

AndyCourcier:
Minor textual changes


manasdk:
Making all these headers appear at the same level is not ideal. I think we will have to look at organizing this doc to fit into sections nicely.

I suggest reverting the header changes and one of us can take a look at those in a separate PR. Keep all the grammar fixes and then I will go ahead and merge this PR :)

Kami:
Thanks for the contribution and I agree with @manasdk on the header changes (this way it's also consistent with other documents).

PR #107, opened 2016-03-30
Remove 'd' from enabled for setting up auth using debs

jjm:
Reported in community by @ogenstad.

The config file has `enable` and the docs say `enabled` so it's a little confusing.


manasdk:
Thanks for the change!

Kami:
Yeah, good catch - thanks :)

PR #108, opened 2016-03-31
Update workflows.rst

heckj:
fix broken doc reference link in published documentation (from https://docs.stackstorm.com/workflows.html)


emedvedev:
Thanks! Good catch.

PR #109, opened 2016-04-01
Add a note about st2auth requirements when using PAM auth backend with the new packages

Kami:
Previously when using AIO installer, the ``st2auth`` process ran as root (yeah, I know, terrible) so this was not an issue unless user didn't use AIO installer and used a custom deployment.

### DIFF
--- a/docs/source/install/__pam_auth_backend_requirements.rst
+++ b/docs/source/install/__pam_auth_backend_requirements.rst
@@ -1,0 +1,8 @@
+.. note::
+
+    When using ``pam`` authentication backend you need to make sure that
+    ``st2auth`` process runs as ``root`` system user otherwise the
+    authentication will fail. For security reasons ``st2auth`` process runs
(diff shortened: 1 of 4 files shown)


docs/source/install/__pam_auth_backend_requirements.rst
```diff
@@ -0,0 +1,8 @@
+.. note::
+
+    When using ``pam`` authentication backend you need to make sure that
+    ``st2auth`` process runs as ``root`` system user otherwise the
+    authentication will fail. For security reasons ``st2auth`` process runs
+    under ``st2`` user by default. If you want to use ``pam`` auth backend and
+    change it to run as ``root``, you can do that by editing service manager
+    file for the ``st2`` auth service.
```
Kami: If someone gets me the service manager file paths for all the supported distros I can also mention this here :P

(don't have access to all the distros right now to look up the exact paths)

PR #110, opened 2016-04-04
Add deprecation notices to the old and out-of-date installation methods

Kami:
Users are keep stumbling on out of date instructions and AIO installer issue and this PR clears things up a bit.

* Remove big red warning from the packages installation pages - package are preferred installation method going forward.
* Add deprecation notice to the "AIO" installer page.
* Add "out of date" notice to the Chef / Puppet / Salt / Ansible installation page
* Drop link to the AIO page in the "Installation" index page

## TODO / To decide

- [x] Remove AIO entry from "Installation" TOC
- [x] Remove Chef / Puppet / Salt / Ansible entry from "Installation" TOC

### DIFF
--- a/docs/source/install/__packages_beta_warning.rst
+++ b/docs/source/install/__packages_beta_warning.rst
@@ -1,7 +1,1 @@
-.. warning::
 
-   Packages based installation is currently is beta and upgrades are not yet
-   supported.
-
-   If you encounter any bugs, please report them at
-   `github.com/StackStorm/st2-packages <https://github.com/StackStorm/st2-packages/issues/new>`_.
--- a/docs/source/install/__packages_beta_warning.rst
+++ b/docs/source/install/__packages_beta_warning.rst
@@ -1,0 +1,7 @@
+.. warning::
+
+   Packages based installation is currently is beta and upgrades are not yet
+   supported.
+
+   If you encounter any bugs, please report them at
+   `github.com/StackStorm/st2-packages <https://github.com/StackStorm/st2-packages/issues/new>`_.
--- a/docs/source/install/all_in_one.rst
+++ b/docs/source/install/all_in_one.rst
@@ -1,5 +1,13 @@
 All-in-one Installer
 ********************
+
+.. note::
+
+    All-in-one Installer has been deprecated and replaced with packages which are now preferred
+    way of installing |st2|.
+    For more more information, please see distribution-specific package based installation
+    instructions - :doc:`deb`, :doc:`rhel6`, :doc:`rhel7`.
+
 |st2| provides an all-in-one installer aimed at assisting users with the initial setup and
 configuration for quick and convenient evaluation.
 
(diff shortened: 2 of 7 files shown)


docs/source/install/__packages_beta_warning.rst
```diff
@@ -0,0 +1,7 @@
+.. warning::
+
+   Packages based installation is currently is beta and upgrades are not yet
```
Kami: I'm open to wording this differently or removing this notice all together.

The fact is that AIO is out of date and not supported so we need to point all the new users to the package based installation instructions and this warning might scare some people off...

emedvedev: I really think we should just remove the outdated methods.

Kami:
If everyone agrees with the proposed changes, we should also merge and publish those docs changes to v1.3.2 since we don't have an exact ETA for v1.4.0 yet.

docs/source/install/__packages_beta_warning.rst
Kami: I also think we should simply remove AIO and Chef / Ansible / Puppet installation documents.

armab:
> Remove big red warning from the packages installation pages - package are preferred installation method going forward.

Side note, removing big red `BETA` warning makes sense only with replacing PackageCloud `staging` urls to production urls (don't know if CI is ready or not).

manasdk:
> If everyone agrees with the proposed changes, we should also merge and publish those docs changes to v1.3 since we don't have an exact ETA for v1.4.0 yet.

Until we have CI or have fully tested community + enterprise installs doing so would be calling for trouble.

Kami:
@manasdk I don't think it really makes a difference.

AIO is broken and we are aware of it and not fixing and supporting it anymore. So based on those facts, only sensible way is dropping all those unsupported (and broken) installation methods.

With the packages we are at least moving forward and fixing issues which pop up...

armab:
Another reminder about this issue from people complaining that `AIO` installer doesn't work: 
https://github.com/StackStorm/ansible-st2/issues/40#issuecomment-207395922

Some action should be made for sure, because `AIO` doesn't work and at the same time Packages are in `BETA`.
It means currently there is no "stable" way to install StackStorm.

docs/source/install/__packages_beta_warning.rst
armab: > `Packages based installation is currently is beta and upgrades are not yet supported.`

This is confusing for me because it's not clear enough what exactly means `upgrades`?
* `upgrades` within new packages for example from `1.3.2` to `1.4.0`? (no problem with that)
* `upgrades` from `AIO` installer to new packages? (it's not clear from that message at all)

Also in case of upgrading from `AIO` to new packages - I think we shouldn't include any info about that at all, because it's just different installation method.
Just big `BETA` notice, +1 with @emedvedev

Kami:
@armab Yeah, that what I was trying to explain to @dzimine.

That's why I wanted to just remove AIO docs since we are not updating or supporting that stuff anymore. With the packages, while they might not be fully ready yet, at least we are supporting it and working on fixing outstanding issues.

PR #115, opened 2016-04-18
corrected mispelling

dankolbrs:
Corrected super minor misspelling in RHEL install documentation.


emedvedev:
Good catch, thanks :)

PR #117, opened 2016-04-18
Add some docs on running / using StackStorm behind an HTTP proxy

Kami:
This issue is coming up again and again and we don't have any docs on it.


docs/source/index.rst
```diff
@@ -28,6 +28,7 @@ Contents:
     rbac
     reference/index
     troubleshooting/index
+    proxy
```
Kami: I'm open to suggestions for a better position in the right navigation bar.

PR #124, opened 2016-04-22
Point docs to production stable repos

lakshmi-kannan:


manasdk:
+1

armab:
From what I see, in [`stable` repo](https://packagecloud.io/StackStorm/stable) there are packages only for:
- `ubuntu trusty`
- `el6`
- `el7`

so `debian wheezy` and `debian jessie` are not promoted (missed?).

^^ FYI

PR #127, opened 2016-04-26
Fixed v1.4 typos

LindsayHill:
Same set of typos as for master. 
Also set copyright date to 2016, and removed AIO reference from Installing Enterprise


emedvedev:
Nice, merging.

PR #131, opened 2016-04-27
Add info about installing client libraries for tooz backends

lakshmi-kannan:


docs/source/policies.rst
```diff
@@ -57,6 +57,16 @@ Configuration example for Redis. ::
...
 
+Some of these coordination backends also require corresponding client libraries to be installed
+in |st2| virtualenv. We do not ship these libraries by default. As an example, to install the client
+library in |st2| virtualenv, run
+
+.. sourcecode:: bash
+
+    sudo su
+    source /opt/stackstorm/st2/bin/activate
+    pip install redis
```
manasdk: Should we recommends a version?

manasdk:
+1

docs/source/policies.rst
lakshmi-kannan: I have no idea what we would recommend because we don't CI this part actively.

Kami:
:+1: 

As mentioned the other day, I think it would also be good to document how virtualenvs are structured and used in package based installations. This will help users understand how it works and make it easier to debug issues, etc.

docs/source/policies.rst
```diff
@@ -57,6 +57,16 @@ Configuration example for Redis. ::
...
     url = redis://password@host:port
 
+Some of these coordination backends also require corresponding client libraries to be installed
+in |st2| virtualenv. We do not ship these libraries by default. As an example, to install the client
+library in |st2| virtualenv, run
+
+.. sourcecode:: bash
+
+    sudo su
+    source /opt/stackstorm/st2/bin/activate
```
Kami: Also, just to make sure permissions are correct, is root the right user to run this command as, or should it be `st2`?

lakshmi-kannan: drwxr-xr-x 7 root root     4096 Jun 22 19:19 st2/

root.

lakshmi-kannan:
>As mentioned the other day, I think it would also be good to document how virtualenvs are structured and used in package based installations.

Good point. We can add this docs as part of 1.5. I'll merge this PR first and work on that next.

PR #144, opened 2016-05-03
[DON'T MERGE] v1.4

Kami:
I just opened this PR so we can see which changes are in v1.4 branch and not in master and vice-versa. This PR shouldn't be merged.

Right now it looks like that we missed some changes in master aka they were merged into v1.4, but not in master.

Having this PR open will allow us to see those missed changes easier and rectify this situation by also porting those changes to master.


Kami:
Actually, I will close this since because of the way we cherry-pick changes onto v1.4 it doesn't work as well as doing `git checkout v1.4 ; git diff master` locally.

PR #161, opened 2016-05-22
Make the AIO deprecation notice more visible

emedvedev:
Because, believe it or not, it’s still not clear enough.

### DIFF
--- a/docs/source/install/all_in_one.rst
+++ b/docs/source/install/all_in_one.rst
@@ -3,9 +3,9 @@
 All-in-one Installer
 ********************
 
-.. note::
+.. warning::
 
-    All-in-one Installer has been deprecated and replaced with packages which are now preferred
+    All-in-one Installer has been **deprecated** and replaced with packages which are now preferred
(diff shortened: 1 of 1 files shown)


Kami:
LGTM, +1

We could also get fully rid of this page (or at least get rid everything exception the warning - this way there is nothing to accidentally follow :P)

In any case, it's something we also need to cherry pick onto v1.4.

Kami:
After seeing that people still some how manage to find those instructions and follow them, I would suggest to remove all the instructions from the page and just leave a warning / notice there with a link to the new package based installation.

emedvedev:
Yeah, I think we'll have to remove old pages at some point, we still have snapshots of previous versions, so it should be fine.

Something else I've seen a lot is a large "THIS IS NOT THE MOST RECENT VERSION" notice when you view outdated docs. This way you can go to the latest version if you just stumble upon a docs page in Google or something.

Anyway, I'll merge this PR for now, and cherry-pick it into 1.4, but that's something to think about.

PR #163, opened 2016-05-24
Docs for pack configuration v2

Kami:
Documentation for https://github.com/StackStorm/st2/pull/2708.


manasdk:
Overall the behavior makes sense to me. Couple of points -

1. A few lines describing the benefit of separating config file from the pack would help put this feature in perspective.
2. Would templatized instad of dynamic capture the idea better? It sort refers to the implementation so helps clarify the behavior.

Kami:
@manasdk 

1. Yeah, I was trying to convey that, but probably also need to make it more explicit (maybe a separate section).

2. Need to think about it some more.

3.. As mentioned in the other PR, I think I also need to add a section on how to migrate from old configs to the new ones, but I'm not sure where I should put that.

Do you think putting it in upgrade notes and then linking to upgrade notes from this document is fine?

Kami:
I will go ahead and merge initial version of docs into master so I can link users to it and start gathering feedback (and improve docs along the way).

PR #169, opened 2016-05-30
Remove AIO docs, leave pointer to new packages

Kami:
People still some how managed to find and follow those deprecated instructions.

As mentioned in https://github.com/StackStorm/st2docs/pull/161#issuecomment-220821139, I'm removing the docs so there is no way to "accidentally" follow obsolete instructions.

Will cherry pick into v1.4.


Kami:
Ideally, we would also update old installer script to print a message which points to the new packages docs and exits.

Kami:
There we go - https://github.com/StackStorm/st2workroom/pull/352

PR #172, opened 2016-06-03
Point vagrant boxes to bento boxes

lakshmi-kannan:
We use bento boxes to do the actual testing! 
https://github.com/StackStorm/st2-test-ground/blob/master/Vagrantfile

So let's be consistent.


armab:
As said before, I'd recommend to use **official** base boxes over any custom boxes where possible.

You want to imitate clean untouched base box as much as possible, while custom boxes provide custom set of software installed, what will bring issues for sure.

```sh
ubuntu/trusty64
centos/7
```

PR #175, opened 2016-06-09
Fix API endpoint url

dzimine:


docs/source/webhooks.rst
```diff
@@ -71,11 +71,11 @@ The body of this request needs to be JSON and must contain the following attribu
 * ``payload`` - Object with a trigger payload.
 
 This example shows how to send data to the generic webhook using
-cURL, and how to match this data using rule criteria:
+cURL, and how to match this data using rule criteria (replace localhost with your st2 host if fire remotely):
```
lakshmi-kannan: Please use emphasis on `localhost` since it has a special meaning.

PR #179, opened 2016-06-14
RFR: Add info about REST API access

lakshmi-kannan:


manasdk:
I don't think it should be in install docs. Only `rest_api_access.rst` is fine in my opinion.

lakshmi-kannan:
I don't think there is any harm in repeating. Merging!

manasdk:
> I don't think there is any harm in repeating. Merging!

Keeping things in sync when you repeat is hard. Anyway you already merged so whatever.

PR #183, opened 2016-06-16
Fixed missing " in Defaults requiretty sed

metareason:


Kami:
Sorry for the delay.

Good catch & thanks!

PR #193, opened 2016-06-22
V1.5 upgrade notes

lakshmi-kannan:
Cherry pick https://github.com/StackStorm/st2docs/pull/192

### DIFF
--- a/docs/source/install/migration.rst
+++ b/docs/source/install/migration.rst
@@ -1,0 +1,26 @@
+.. _migration-scripts-to-run:
+
+Migration scripts to run on upgrades
+====================================
+
+Migration scripts most often need to be run *before* upgrading |st2| packages.
+The runbook is usually
+
+1. Run the migration script
+2. Upgrade ``st2`` packages using distro specific tools (
+``apt-get install --only-upgrade $PKG_NAME`` for debian based and ``yum update $PKG_NAME``)
+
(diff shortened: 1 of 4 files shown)


docs/source/install/migration.rst
```diff
@@ -0,0 +1,29 @@
...
+respective version
+
+v1.5
+----
+
+* Datastore model migration
+
+::
+
+    /opt/stackstorm/st2/bin/st2-migrate-datastore-to-include-scope-secret.py
```
armab: The good way is to benefit from packaging and make this process automated during packages upgrades.

Otherwise you're back with one-legged solutions which make your users suffer again (like it was with old packaging).

armab:
Was the `sudo: true` parameter removed for `local-shell-script` in `v1.5`?

docs/source/install/migration.rst
manasdk: Yes, running it as part of install is ideal. Hopefully, we get to it in the next release.

lakshmi-kannan: @armab Yep, make it part of packaging was my first attempt. However, our post install might grow like crazy if we did multiple migrations in the future. So ideally, we ship migration scripts and an uber migration script that checks versions and invokes the appropriate migration scripts. We are targeting this for next release.

armab: Sounds like a plan :+1:

PR #203, opened 2016-06-28
Minor change to instruction for mistral database migration

m4dcoder:


Kami:
Thanks for this writeup.

Re-building the whole virtual environment seems a bit excessive and shouldn't be needed, right? And unless I am missing something, all the dependencies now ship with st2mistral package (they are bundled) so the virtualenv related stuff shouldn't be necessary and in fact, might break things.

lakshmi-kannan:
+1 to what kami is saying. We shouldn't muck with virtualenv steps with the new packages. It defeats the purpose of packages. I am fine with other steps other than the whole virtualenv part. I am not sure if mistral services need to be stopped to run the migration scripts. If indeed it is required, then leave the stop and start. 

Also, need to push changes to v1.5 branch as well.

docs/source/install/migration.rst
```diff
@@ -31,13 +31,24 @@ RHEL / CentOS:
...
 
-4. Run Mistral database migration scripts:
+4. Upgrade Mistral database.
 
 .. sourcecode:: bash
 
+  # Stop related services
+  service mistral-api stop
+  service mistral stop
+  service postgresql stop
```
Kami: Hm, I believe PostgreSQL should be running here, otherwise migration will fail, right (since migration populates db)?

Issue #201, opened 2016-06-27
References to gist.github.com


manasdk:
https://github.com/StackStorm/st2docs/blob/ed3fa0c113d1d0681c7e6e4a46cc91733bfa9dea/docs/source/reference/ha.rst contains references to https://gist.github.com/manasdk/fce14029900e533a385d. I don't plan to delete these anytime soon but perhaps a better idea to take over the content. Perhaps put it under a StackStorm owned and managed github users gist.

m4dcoder:
https://github.com/StackStorm/st2docs/pull/204

PR #204, opened 2016-06-28
Remove gist references from the high availability page

m4dcoder:


m4dcoder:
RE: https://github.com/StackStorm/st2docs/issues/201

docs/source/reference/ha_shared_st2_nginx.conf
```diff
@@ -0,0 +1,114 @@
+#
```
lakshmi-kannan: Can you please save these files in https://github.com/StackStorm/st2/tree/master/conf/HA/ and then point to them? Otherwise, we might end up introducing a drift in config in the long run.

lakshmi-kannan:
LGTM. Merge to v1.5 as well.

PR #212, opened 2016-07-14
Added Mac instructions

pete911:


emedvedev:
Last time I checked, StackStorm on Mac failed because there's no `inotify`. Are you able to launch it successfully?

On a related note, the preferred method to run St2 on a home machine or a laptop is a Vagrant image, and I would advise against installing St2 without a VM.

pete911:
Hi, yeah, I just noticed that after I made this pull request. I was able to work on it (haven't tried to run it yet though) after commenting out lines 7 and 26 in requirements.txt

emedvedev:
Unfortunately, those lines are there for a reason, and it's possible you won't be able to properly run StackStorm without `inotify`. :) I'd really recommend installing it inside an Ubuntu / CentOS VM.

pete911:
yeah, this was for development so I can load it to pycharm, not for running. But didn't know which part of documentation to update. Can you point me to 'development' part of documentation, I can add it there instead

PR #213, opened 2016-07-16
Update documentation for MongoDB 3.2 changes, reduce duplication / copy & paste

Kami:
This pull request updates documentation for MongoDB 3.2 changes.

In addition to that, I have finally done some clean-up and refactoring and moved duplicated installation steps which are the same across all the platforms in shared files (something I have complained a lot about in the past).

Keep in mind though that this is just a very first step. Next step will actually be directly re-using code from bootstrap scripts. I strongly believe that documentation which is not automated and tested (like it is right now) is bad and in many cases even worse then no documentation (things get out of date, etc.)


LindsayHill:
Nice, I like the approach of including common blocks for install guides. Should mean better consistency in future

Kami:
Yeah, still needs a bit more refactoring love, but I lost motivation so it will be done at some point in the future :)

PR #224, opened 2016-08-02
Action and runner doc update for python_action_runner

humblearner:


Kami:
Couple of typos I will fix in master.

Besides that, LGTM :+1:

docs/source/actions.rst
```diff
@@ -522,11 +525,13 @@ Action script file (``my_echo_action.py``):
 
     class MyEchoAction(Action):
         def run(self, message):
-            print(message)
-            sys.exit(0)
+            if message == 'working'
+                return (True, message)
```
Kami: You changed the behavior so this needs to be reflected in the code as well.

Action now returns a result which is not the same as printing a message (previous behavior). I will update this in master.

PR #231, opened 2016-08-09
Repeat after me, 64 bit is the only real thing

lakshmi-kannan:


docs/source/install/index.rst
```diff
@@ -3,15 +3,15 @@ Installation
...
+Please note that only 64 bit architecture is currently supported.
 
 .. rubric:: Installations
 
 .. toctree::
     :maxdepth: 1
 
     Reference Deployment <overview>
-    Ubuntu / Debian <deb>
+    Ubuntu / Debian 14.04 (Trusty) <deb>
```
armab: There is also 16.04, thanks to @enykeev efforts.
So I think it's better how it is.

Apart of that, 
https://docs.stackstorm.com/install/deb.html#supported-versions
> We support Ubuntu 14.04, and test on Ubuntu Server 14.04 LTS (HVM) Amazon AWS AMI and puppetlabs/ubuntu-14.04-64-nocm Vagrant box. Other Debian based distributions and versions will likely work with some tweaks. You are welcome to try - please report success to the community.

So that's really enough.
Repeating one thing 3 times for people who can't hear won't help.

lakshmi-kannan: Yep. I know about 16.04. When we have CI, I'll edit the docs and mention 16.04. The supported versions doesn't talk about 64 bit and unfortunately a lot of people try 32 bit because Ubuntu downloads are confusing (they call their 64 bit ISOs amd-64). We decided to explicitly call out 64 bit in docs but I am the one making the actual changes

armab: This line just doesn't look nice now :)

BTW, for those who is OK not to get support, there are debian `wheezy` and `jessie` packages in unstable repo. I know some people are using it.
What do you think about the future of those packages?

lakshmi-kannan: >there are debian wheezy and jessie packages in unstable repo

Here's a fun problem. I basically don't want to claim support unless it goes through our CI machinery. If you are on #stackstorm, you'd see atrocious numbers of CI failures just for 3 OS for which we CI regularly. Most of the times, it is a 3rd party thing we depend on - GNU TLS package cloud, AWS APIs timing out etc. There is a huge pain now to make sure 3 OSes get CIed correctly. If we add more to the mix, we'll essentially kill ourselves with just debugging build failures. So I am extremely skeptical about adding more OSes. From a customers' point, no one has asked for those. Many enterprises are switching to ubuntu server images if not using them already.

humblearner:
:D

docs/source/install/index.rst
armab: Can't say I'm happy with that.

But at least it's OK for me, since this information appeared on horizon: https://github.com/StackStorm/st2-dockerfiles/issues/23 meaning

armab: That line :+1: Now it looks better, comparing to old `... Debian 14.04`

lakshmi-kannan: That was me being dumb like usual :). Good catch.

enykeev: purely semantically, what debian is doing in this line at all? There's surely no `debian 14.04` in the wild. I'd rather switch the order.

Kami:
I will also cherry pick this into v1.6 and add note banner since people still miss it.

PR #233, opened 2016-08-09
Chatops alias example to use simple remote shell cmd

lakshmi-kannan:


emedvedev:
Do not merge just yet, have small comments, will expand later (on my way to the airport now).

docs/source/chatops/aliases.rst
```diff
@@ -103,11 +104,12 @@ Using this example:
...
-      - "google {{query=StackStorm}}"
+      - "run {{cmd}} on {{hosts=localhost}}"
 
-In this case the query has a default value assigned which will be used if no value is provided by the user.
-Therefore, a simple ``google`` instead of ``google StackStorm`` would result in assigning the
-default value, in a similar manner to how Action default parameter values are interpreted.
+In this case the query has a default value assigned which will be used
+if no value is provided by the user. Therefore, a simple ``run date`` instead of
+``run date on localhost`` would result in assigning the default value, in a similar
+manner to how Action default parameter values are interpreted.
```
emedvedev: > Therefore, a simple `run date` instead of `run date on localhost` would result in assigning the default value

It won't. `run date on` will, because the "on" part is not optional. :) I'll change it to something slightly different.

emedvedev:
Fixed a couple discrepancies, expanded on the list of values, and updated `chatops.rst`, since it still referenced the Google example.

Everything else LGTM. Take a look through my changes and merge.

lakshmi-kannan:
Looks as perfect as a little kid :P. I'll merge when travis blesses.

PR #237, opened 2016-08-17
Move StackStorm/BWC-specific values from conf.py to info.py

emedvedev:
This is needed for BWC docs build: we'll just have a different `info.py` there, while `conf.py` will be a single source of truth, and there'll be no need to constantly keep a bunch of config files in sync.


DoriftoShoes:
+1

PR #244, opened 2016-08-19
Make local bwc docs

DoriftoShoes:


emedvedev:
Will do for now. :+1:

PR #246, opened 2016-08-19
RFR: Have two different targets for enteprise and community docs

lakshmi-kannan:
## What?

This PR uses sphinx tags to selectively include different sections for community docs and enterprise docs. 

## Makefile targets

```
make .docs -> make .community-docs
make .community-docs
make .enterprise-docs
```

## How to use tags?

See example https://github.com/StackStorm/st2docs/pull/246/files#diff-a49a378abfac4a4c0ffbbc19fc8732b5R14


emedvedev:
It would be cleaner (and easier on our CI) to make an "enterprise" flag in `conf.py`/`info.py`, set it to `True` in BWC docs, and then use (ifconfig)[http://www.sphinx-doc.org/en/stable/ext/ifconfig.html]:

```
.. ifconfig:: enterprise is True

   This is an enterprise-only block.
```

I'd prefer this. If you don't want to change stuff, we can yoloship. :shipit:

lakshmi-kannan:
>It would be cleaner (and easier on our CI) to make an "enterprise" flag in conf.py/info.py, set it to

I first went that route. You can do tags.add('enterprise') in conf.py. It's allowed but I didn't like it. We anyways have to have separate make targets.

dzimine:
LGTM.
Nit: I'd keep `doc` for brevity / as default, and use `make docs` and `make enterprise docs`. But code wins, you have it your way @lakshmi-kannan !

PR #247, opened 2016-08-21
Changed Flow to Design UI

ramgolla:


emedvedev:
Where is this change coming from? I don't remember anything about renaming Flow.

If it's already decided that we should rename it, then there's a bunch of mentions everywhere else in the docs, so we should rename it everywhere at once.

cc @dzimine

ramgolla:
Hi Ed,

Yes, Flow will be called Design UI.  But before major changes to the docs we will outline the agreed category of changes on Slack tomorrow for everyone to review before making the changes to the docs.
This change I made just to understand the GitHub – leave it for now.

Thanks,
Ram

dzimine:
https://stackstorm.atlassian.net/browse/STORM-2269
Flow is Brocade Workflow Designer, last time I heard about it.

PR #248, opened 2016-08-22
Build and run docs with docker

dzimine:
I got worried about explaining "how to write the docs" to less-technical folks. 

Using Docker for live-editing the docs, and pushing the doc-only PRs seem to be simpler and faster. Especially for those who have no st2 dev environment. With docker coming natively on Windows and Mac, it may be even simpler setup for windows users. Let's test on PMs :)

Once Docker is installed: 
```bash
git clone https://github.com/StackStorm/st2docs.git
cd st2docs
./scripts/docker-build.sh  # Builds an image with all the dependencies, runs 10+ min, grab a coffee
./scripts/docker-run.sh # Runs the doc site live. Enjoy.
```

That's it. Edit the sources and enjoy live updates at [http://localhost:8000](http://localhost:8000)

### DIFF
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+pushd `dirname $0` > /dev/null
+ST2DOCSROOT="${PWD%/*}"
+popd > /dev/null
+
+docker build -t st2/st2docs -f $ST2DOCSROOT/Dockerfile $ST2DOCSROOT
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+virtualenv
+st2
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu
+
+RUN apt-get -qq update && apt-get -q install -y \
+    git \
+    python-dev python-pip python-virtualenv \
+    libffi-dev libssl-dev
+
+ADD . /st2docs
+WORKDIR /st2docs
+RUN make .cleandocs && make docs
+
+EXPOSE 8000
+
+CMD make .livedocs
+
+#CMD  . ./virtualenv/bin/activate; \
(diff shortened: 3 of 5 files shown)


LindsayHill:
Thanks for this. Might help with a problem I was having with building a BWC docs environment on a Mac. Might also need to tweak it a bit to allow for building a live BWC docs env (equivalent of make bwclivedocs)

scripts/docker-build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+pushd `dirname $0` > /dev/null
+ST2DOCSROOT="${PWD%/*}"
+popd > /dev/null
+
+docker build -t st2/st2docs -f $ST2DOCSROOT/Dockerfile $ST2DOCSROOT
```
armab: More obvious thing looks like to include the instructions in a `Makefile`, so the commands would be: 

```sh
make docker-build
make docker-run
```

armab:
Nice one :+1: 

We can also deploy docker images to https://hub.docker.com/r/stackstorm/ to avoid `docker-build.sh` and save `10mins`.

scripts/docker-build.sh
dzimine: I afraid it gets `meta`: I use `make` from docker...  So will keep it a bit separate for now.

While doing it I saw that so many things can be off-loaded from Makefile if docker was the way of building... But taking it simple for now.

dzimine: @LindsayHill absolutely possible with something like this: 

```
+docker run --rm -it -p 127.0.0.1:8000:8000 \
 +  -v "$ST2DOCSROOT"/docs/source:/st2docs/docs/source st2/st2docs  /bin/bash -c "make .cleandocs; make bwclivedocs"
```
We can add wrapping scripts, but I guess @lakshmi-kannan is deliberately keep all BWC (non-community stuff) existing, but not documented on community doc repo.

PR #249, opened 2016-08-22
bwclocaldocs ain't need to build no docs

lakshmi-kannan:
I forgot to remove .docs target from bwclocaldocs. Now we build enterprise docs separately and the target for that is .enterprise-docs which is already called.


humblearner:
LGTM

DoriftoShoes:
+1

PR #250, opened 2016-08-23
Makefile update to remove solutions stuff and consolidate solutions images

humblearner:
Refer: https://github.com/StackStorm/ipfabric-docs/pull/24

- All the images in the solutions should be placed under _static/images/solutions/<solution_name>/ to keep the make file in st2docs clean.
- Add `.remove-bwc-solutions-folders` target to remove any solutions docs before making `community-docs` and to clean up after `bwclocaldocs`. ( `make docs` was failing, if make targets for bwc were executed before)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,10 +48,10 @@ docs: .clone-st2 requirements .requirements-st2 .docs
 PHONY: .docs
 .docs: .community-docs
 
-PHONY: .remove-bwc-solutions-folders
-.remove-bwc-solutions-folders:
+PHONY: .clean-bwc-solutions-folders
+.clean-bwc-solutions-folders:
 	@echo
-	@echo "==================== Remove solution specific stuff ===================="
+	@echo "==================== Remove solution specific folders ===================="
 	@echo
 	rm -Rf docs/source/_static/images/solutions/ipfabric/*
 	rm -Rf docs/source/_includes/bwc_toc.rst
@@ -60,7 +60,7 @@ PHONY: .remove-bwc-solutions-folders
 	@echo
 
 .PHONY: .community-docs
-.community-docs: .remove-bwc-solutions-folders
+.community-docs: .clean-bwc-solutions-folders
 	@echo
 	@echo "==================== COMMUNITY DOCS ===================="
 	@echo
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -88,7 +99,7 @@ livedocs: docs .livedocs
 	@echo
 
 .PHONY: bwcdocs
-bwcdocs: .clone-st2 .clone-ipfabric requirements .requirements-st2 .patch-solutions .enterprise-docs .git-checkout-local-changes
+bwcdocs: .clone-st2 .clone-ipfabric requirements .requirements-st2 .bwcdocs
```
lakshmi-kannan: Good catch!

Makefile
```diff
@@ -48,8 +48,19 @@ docs: .clone-st2 requirements .requirements-st2 .docs
 PHONY: .docs
 .docs: .community-docs
 
+PHONY: .remove-bwc-solutions-folders
```
lakshmi-kannan: Nit: Usually these targets are called "clean" targets in Makefile lingo. So .clean-bwc-solutions-folders might be a better name.

Makefile
```diff
@@ -48,8 +48,19 @@ docs: .clone-st2 requirements .requirements-st2 .docs
...
+.remove-bwc-solutions-folders:
+	@echo
+	@echo "==================== Remove solution specific stuff ===================="
```
lakshmi-kannan: Lol. Your obsession with the word "stuff" is unhealthy.

Makefile
emedvedev: Yeah, let's do that first.

dzimine:
+1 for this PR; while I'd prefer a shorter more expressive phony like clean-solutions, or `unpatch-solutions` if this is symmetric to `.patch-solutions` in what it does. 

Good naming scheme would have helped; else - we copy * when patch, but unpatching requires knowledge of ipfabric-docs structure, and when it changes, the files will change in two places. With a naming convention it may be `rm -rf source/_include/bwc*`.

Catching up: why are we unpatching the docs with `.git-checkout-local-changes`? 
Looks like to keep source "as it was" after the `make bwcdocs`, so that `make docs` can run? 
It's a bit confusing to me to run `make bwcdocs`, than `make .livedocs`, and see that init.py is from the community, not from BWC as I would have expected. I understand that the docs was generated with a right one but still weird. May be I just need to understand why it is needed to un-patch.

Makefile
```diff
@@ -48,8 +48,19 @@ docs: .clone-st2 requirements .requirements-st2 .docs
...
+	@echo
+	rm -Rf docs/source/_static/images/solutions/ipfabric/*
```
dzimine: should we ditch all solutions? `rm -Rf docs/source/static/images/solutions/*` ?

dzimine:
+1 for this PR; while I'd prefer a shorter more expressive phony like clean-solutions, or `unpatch-solutions` if this is symmetric to `.patch-solutions` in what it does. 

Good naming scheme would have helped to avoid changing here when ipfabric-docs changed, it may be `rm source/_include/bwc-*.rst`

Catching up: why are we unpatching the docs with `.git-checkout-local-changes`? 
Looks like to keep source "as it was" after the `make bwcdocs`, so that `make docs` can run?

emedvedev:
I can live with this. +1.

PR #253, opened 2016-08-23
[WIP] Rename Enterprise to Brocade Workflow Composer

yhasan:
This is a "try the water PR"

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,5 +1,5 @@
-Welcome to |st2|'s documentation!
-======================================
+Welcome to BWC documentation!
+==============================
 
 Contents:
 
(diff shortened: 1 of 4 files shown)


docs/source/install/index.rst
```diff
@@ -15,6 +15,6 @@ Please note that only 64 bit architecture is currently supported.
     Ubuntu (Trusty) <deb>
     RHEL 7 / CentOS 7 <rhel7>
     RHEL 6 / CentOS 6 <rhel6>
-    StackStorm Enterprise <enterprise>
+    Brocade Workflow Composer <enterprise>
```
dzimine: in TOC, we need to contrast ST2 with BWC, how can we do this?

docs/source/index.rst
```diff
@@ -1,5 +1,5 @@
-Welcome to |st2|'s documentation!
-======================================
+Welcome to BWC documentation!
```
LindsayHill: I think this one should remain `|st2|`

PR #255, opened 2016-08-24
Add product name switch

dzimine:
Switch BWC and StackStorm product names based on the `enterprise` tag we pass to makefile
for `docs` and `bwcdocs`

Also, tiny clean-up.


docs/source/conf.py
```diff
@@ -114,13 +114,21 @@ def previous_version(ver):
 
 # Inserted at the bottom of all rst files.
 # Use for variables substitutions
+
+if tags.has('enterprise'):
+    print "Building BWC docs"
```
lakshmi-kannan: You want these prints?

Kami: I assume so so it's visible during build...

dzimine: yes, indeed: visible during build. Else I need to look at results, and if I'm on travis, or build003, good to check

PR #256, opened 2016-08-25
[WIP] Reorganize TOC

dzimine:
Rational: simplify the view, hide crap, provide better visibility and navigation to fundamentals,  at the expense of navigations for less used / more advanced items.
Pros: Clean first level navigation.
Cons: everything in `References and Guidelines` is a mess. It's a content problem - a mixture of references and guides and tutorials and  notes... Not fixing it now. 

TODO: 
- [x] New structure
- [x] Validate  and fix all links
- [x] Add styles for TOC sections (`caption` tag).
- [ ] Adjust text where required for smoother flow after reorg.

How it looks (with `bwddocs`): 
pasted image at 2016_08_25 01_32


dzimine:
and this is the sections with community skin: 
image

docs/source/_themes/sphinx_rtd_theme/static/css/theme.css
```diff
@@ -3488,7 +3488,7 @@ a.wy-text-neutral:hover {
   color: #595959 !important;
 }
 
-h1, h2, h3, h4, h5, h6, legend {
+h1, h2, h3, h4, h5, h6, .rst-content .toctree-wrapper p.caption, legend {
```
enykeev: This looks like reset styles, they suppose to make browser implementations for particular elements look the same. I'd rather you copy font's weight and family to specific selector you already have down below. Not a deal breaker, but slightly more preferred way.

docs/source/_themes/sphinx_rtd_theme/static/css/theme.css
```diff
@@ -3525,6 +3525,10 @@ h6 {
   font-size: 100%;
 }
 
+.rst-content .toctree-wrapper p.caption {
```
enykeev: Unless you have other elements with class "caption" laying around, `p` is not needed here. One again, minor as it gets.

docs/source/_themes/sphinx_rtd_theme/static/css/theme.css
dzimine: copied from new original read_the_docs theme. I can do this, too.

LindsayHill:
Anyone else seeing a carriage return symbol at the end of each 'caption' entry in index.html?

LindsayHill:
For the record, I am totally +1 on the idea of having these sections. Need to haggle over exact sections we use. I am wondering if we need to break it up into more than just Fundamentals & Advanced. 

I also think that Network Automation Suites should appear above the "Advanced" section if you're looking at BWC docs.

LindsayHill:
What do you think of this TOC structure?
- Getting Started
  - Overview
  - Installation
    - Reference Deployment
    - StackStorm vs BWC
    - (Various OS flavours)
    - Upgrading
  - Quick Start

- Using ST2/BWC
  - Integration Packs
  - Actions
  - Rules
  - Workflows
  - Triggers and Sensors
  - Webhooks
  - Datastore
  - Authentication
  - RBAC
  - ChatOps

- Advanced Topics
  - References
    - CLI
    - Pack Configuration
    - Pack Creation/Contribution
    - Jinja
    - Policies
    - Action Runners
    - Traces
    - Libaries and SDKs
    - Using BWC behind a Proxy
    - HA Deployment
    - API Reference
    - Partitioning Sensors
    - History and Audit
  - Troubleshooting
  	- (Various troubleshooting sub-topics)
  - Development/Contributing

- Release Notes
  - Changelog
  - Upgrade Notes
  - Roadmap

emedvedev:
I think what @LindsayHill proposed is nice.

ramgolla:
What happened to the Network Automation Suites in the last rev?  I think it should be in the 1st tier topic.

LindsayHill:
@ramgolla Automation Suites only appear in the BWC version of the docs. I didn't include it in my proposal above, but I think it would fit in just before "Advanced Topics"

dzimine:
I am going with @LindsayHill suggestion. I'll be creative if something doesn't fit exactly but follow as close as I can. 

Thanks!

docs/source/_themes/sphinx_rtd_theme/static/css/theme.css
dzimine: this didn't work and I didn't go after figuring out why, css is massive. Leaving as is.

dzimine:
Bummer... substitutions don't work with directive parameters, 
so this doesn't work: 

```
.. toctree::
    :maxdepth: 2
    :caption: Using |fullname|
```
and "Fundamentals" look a bit terroristish for a label. 

I move on push what I have so far, and return to playing with this with conditions later testing.  Meantime, if anyone finds a good section name that doesn't require substition, bring it up!

PR #257, opened 2016-08-25
RFR: added |fullname| substitution

LindsayHill:
We have a substitution that maps `|st2|` to 'StackStorm' for the community docs, and to 'BWC' if the docs are built with `-t enterprise`

We also have a substitution that maps `|bwc|` to `Brocade Workflow Composer`.

But sometimes we need to map `|st2|` to either `StackStorm` or `Brocade Workflow Composer` - e.g. for /index.html, we want to say "Welcome to Brocade Workflow Composer's Documentation!"

This change is a little bit ugly. Adds a |fullname| substitution. 

Thoughts @humblearner ?

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -117,10 +117,10 @@ extlinks = {
 
 if tags.has('enterprise'):
     print "Building BWC docs"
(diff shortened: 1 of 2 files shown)


humblearner:
LGTM

PR #258, opened 2016-08-26
added verbiage around enterprise key

LindsayHill:
Better explanation of how to use the license key, based upon feedback from @Plaporte


LindsayHill:
NB: Names will change in this section (Enterprise -> BWC). Didn't want to make those changes as part of this.

PR #265, opened 2016-08-30
disable pilcrow symbol

LindsayHill:
This disables the permalink and the pilcrow (¶) symbol on the Caption headings in the TOC.

Seems to still work for displaying permalinks in normal pages - they have a symbol that only appears on hover.


StorminStanley:
Interesting. Keep guessing who wrote this!

LindsayHill:
OK, first attempted fix was wrong way. Have now made change in theme.css. This works as expected, but I'm not 100% certain it's the Right Way to solve it. 

@enykeev could you please take a look? I don't want to merge until you've had a chance to check it. If it is OK, we'll probably also need to update the theme.css used for ipfabric docs

PR #267, opened 2016-08-31
Renaming, Heading Cleanup, TOC Changes

LindsayHill:
Went through naming, explicitly using StackStorm in some locations where more appropriate, in others switching from StackStorm to |st2|

Modified some TOC depths where they were too deep - e.g. main index.html didn't really need to show an entry for every individual changelog entry. 

Updated lots of headings. Some of them changed the heading level so that we dont get unnecessary things showing up in TOCS.  Lots of capitalisation changes in headings too, to get some consistency. Not perfect yet.

Various small typo & language changes.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -78,7 +78,7 @@ actions to be run remotely (via SSH) and locally. The objective is to
 allow the Action author to concentrate only on the implementation of the
 action itself rather than setting up the environment.
 
-Available runners
+Available Runners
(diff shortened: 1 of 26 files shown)


lakshmi-kannan:
LGTM.

dzimine:
That's great deal of work. Thank you! 

Note: we will need to cherry-pick it to v2.0. I suggest we do it all at once, after we fix it all on master to our liking.

PR #269, opened 2016-08-31
Fixing some grossly outdated parts of the doc

enykeev:


docs/source/install/config/config.rst
```diff
@@ -6,7 +6,7 @@ serving static content of WebUI,
 and running st2auth and st2api as WSGI apps via gunicorn/uwsgi. |st2| nginx configurations
 can be found at ``/etc/nginx/sites-enabled/st2*.conf``.
 
-``st2auth`` and ``st2api`` can also run using a built-in simple Python server. This is used for development and strongly discouraged for any production. Be aware that some settings in /etc/st2.conf are only effective when running in development mode, and don't apply when running under WSGI servers. Refer to the comments in
+Be aware that some settings in /etc/st2.conf are only effective when running in development mode, and don't apply when running under WSGI servers. Refer to the comments in
```
Kami: Should this also be /etc/st2/st2.conf?

And should make it clear that this only applies to listen ip and port settings for st2 auth and st2 api.

enykeev:
- “But there is no joy without WebUI, no security without SSL termination, no fun without ChatOps, and no money without Brocade Workflow Composer.” pretty useless and boring project, ei!
- Outdated upgrade notice in https://bwc-docs.brocade.com/latest/install/deb.html#upgrade-to-brocade-workflow-composer. Should also mention `install.sh` or be discarded completely.
- Changing st2 to bwc created a lot of confusion in parts of the docs that direct you to run the st2 command. When I read something along the lines “to test BWC, run `st2…`” my first reaction is “wait, what? How is this even related?”

dzimine:
+1

PR #270, opened 2016-08-31
Merge two Jinja docs to one.

dzimine:
it says it.


LindsayHill:
Good work. Probably cherry-pick to 2.0? Doesn't matter too much either way

PR #273, opened 2016-09-02
Fix the regex used to disable 'Default requiretty'

jvrplmlmn:
As discussed on Slack, the regex won't work if it finds something like `Defaults   requiretty` (without the `+`).

@Kami ?


estee-tew:
https://stackstorm.slack.com/archives/community/p1472832544000335

Kami:
Great, thanks!

Looks good to me. To clarify it - I also checked the bootstrap script - the scripts itself are fine (they do make `+` optional), but it looks like docs some how got out of sync.

I will merge this into master and also cherry pick it into v2.0 branch.

Thanks again.

PR #274, opened 2016-09-02
Added note about updating curl

LindsayHill:
Make sure curl is updated before starting. Critical for RHEL 6, but also useful for other distros.

### DIFF
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -2,16 +2,18 @@ Installation
 ============
 
 That's OK! You're busy, we get it. How do you just get started? Get yourself a clean 64-bit Linux
-box that fits the :doc:`system requirements <system_requirements>`, and run this command:
+box that fits the :doc:`system requirements <system_requirements>`. Make sure that ``curl`` is up to date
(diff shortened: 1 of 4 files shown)


docs/source/install/index.rst
```diff
@@ -2,16 +2,18 @@ Installation
 ============
 
 That's OK! You're busy, we get it. How do you just get started? Get yourself a clean 64-bit Linux
-box that fits the :doc:`system requirements <system_requirements>`, and run this command:
+box that fits the :doc:`system requirements <system_requirements>`. Make sure that ``curl`` is up to date
+using ``sudo apt-get install curl`` on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run
```
dzimine: [soft] one extra idea is to create a bug in st2, close it as resolved, and have a ref to the bug from here in addition.

dzimine:
+1

docs/source/install/index.rst
emedvedev: It might be nice, but it's probably better not to distract people before the installation even happened. :)

PR #275, opened 2016-09-09
Update libffi-devel url

pengyao:
CentOS 6.8 have been released. It's only REAEME in 6.7

Use 6 which is the latest version  instead release version.


armab:
:+1:

PR #279, opened 2016-09-16
KB Article Parsing

bigmstone:
This PR is to facilitate automatic KB article push to Remaze. It will work in conjunction with an `st2cd` workflow that will take the generated KB articles and upload them to Remaze.

Todo:
- [x] Move initial attempt over to YAML based files and parsing
- [x] Generate RST based on YAML KB article
- [x] KB Header information
- [x] Adding links to KB page to docs
- [x] Add KB generate to makefile
- [x] Topic Sections
- [x] Heading manipulation to honor RST title level consistency
- [ ] st2cd workflow (will link when available)


bigmstone:
Something to note: The metadata portion of the KB payload is only there as filler right now. `metadata: true` is arbitrary and means nothing. I know there will need to be some metadata associated with the KB, but I haven't collected all that information yet.

m4dcoder:
Is it possible to separate FAQ/KB articles in separate files?  I think having KB related articles embedded in current docs can get unwieldy very quickly.

bigmstone:
This will parse any file in the docs repo. So we can separate it out as we wish. It was done that way on purpose so we can move things around as we wish. I will move the example to its own section. Others can comment on that as well.

bigmstone:
Another option is to have a folder full of YAML files that build the FAQ or KB page on the fly before the docs are generated. Perhaps that's what you meant and that was lost on me.

bigmstone:
I tend to favor that solution. That would be much easier to parse.

bigmstone:
Okay. Latest commit adds in the YAML based parsing, etc. Couple more things to tidy up and then we'll be ready to go.

m4dcoder:
Does this work for multiline string using `>` in yaml?

m4dcoder:
Are we expecting the CI jobs to go green? Or there's a reasonable explanation for staying red?

kb/adding-os-support-to-all-in-one-installer.yml
```diff
@@ -0,0 +1,13 @@
...
+  \ your help together the necessary items to run on your OS of choice.\r\n\r\n##\
+  \ Suggested Reading\r\n\r\nhttps://stackstorm.reamaze.com/articles/installer-internals-and-flow\r\
+  \n\r\nWhile not absolutely necessary, it is highly recommended to take a look at\
+  \ the [Installer Internals and Flow](https://stackstorm.reamaze.com/articles/installer-internals-and-flow)\
+  \ article. This shows at least the overview and goals of the Installer, and where\
+  \ it is headed as well as we grow/learn/mature.\r\n\r\n## Things/places to add.\r\
+  \n\r\n> TBD"
+status: 0
+title: Adding OS Support to All-in-one Installer
+topic: Installation
```
m4dcoder: Does it make sense for an article to belong to mutliple topics?

kb/adding-os-support-to-all-in-one-installer.yml
```diff
@@ -0,0 +1,13 @@
...
+  \ and want to take StackStorm along on your voyage. Well, frankly, we're flattered!\
+  \ But, sometimes we need some help getting geared up for new places, and could use\
+  \ your help together the necessary items to run on your OS of choice.\r\n\r\n##\
+  \ Suggested Reading\r\n\r\nhttps://stackstorm.reamaze.com/articles/installer-internals-and-flow\r\
+  \n\r\nWhile not absolutely necessary, it is highly recommended to take a look at\
+  \ the [Installer Internals and Flow](https://stackstorm.reamaze.com/articles/installer-internals-and-flow)\
+  \ article. This shows at least the overview and goals of the Installer, and where\
+  \ it is headed as well as we grow/learn/mature.\r\n\r\n## Things/places to add.\r\
+  \n\r\n> TBD"
+status: 0
```
m4dcoder: What does status means?

scripts/generate-kb-articles.py
```diff
@@ -0,0 +1,91 @@
...
+    body = body.replace('\r**', '**\r')
+    convert_string = "### %s\n %s" % (article['title'], body)
+    return pypandoc.convert_text(convert_string, 'rst', 'md')
+
+
+def write_rst(kb_data):
+    """Write RST file to disk.
+    """
+    with open(RST_PATH, "w+") as rst_file:
+        rst_file.write(RST_HEADER),
```
m4dcoder: There's a comma at the end of this file. Is the `with` state following continuation of this line?  Or is it a separate block after?

scripts/generate-kb-articles.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+
+def get_files():
+    """Get document files and prepare to parse.
+    """
+    doc_files = []
+    for (dirpath, _, files_in_dir) in os.walk(ST2DOCS_PATH):
+        for files in files_in_dir:
+            doc_files.append(dirpath + '/' + files)
```
m4dcoder: I typically prefer spacing before a return.

scripts/generate-kb-articles.py
```diff
@@ -0,0 +1,91 @@
...
+    """Write RST file to disk.
+    """
+    with open(RST_PATH, "w+") as rst_file:
+        rst_file.write(RST_HEADER),
+    with open(RST_PATH, "a") as rst_file:
+        for topic, articles in kb_data.iteritems():
+            rst_file.write(topic + "\n" + "-"*len(topic) + "\n\n")
+            for article in articles:
+                rst_contents = convert_to_rst(article)
+                print article['title']
```
m4dcoder: Should use `()` w/ print to be py3 compatible.

bigmstone:
>Does this work for multiline string using > in yaml?
Yes as well as `|`. The formatting is strange on a majority of these files because of the way pyaml dumps them. Not much I can do there save manually (sed/awk foo) fixing them. See [here](https://github.com/StackStorm/st2docs/blob/be7b2d9c0e04f0440226b241d60eef18e56b5d0d/kb/reset_password.yml) for an example.

>Are we expecting the CI jobs to go green? Or there's a reasonable explanation for staying red?
As far as I can tell this is failing on something unrelated. I assumed master was failing as well.

>What does status means?
It's metadata from reamaze. In this case `status` of 0 means the article was in the `published` state.

>Does it make sense for an article to belong to mutliple topics?
I can conceive of how that could be useful, but I am not the best person to answer that. It does beg the question of how to organize them on a page if they are a part of multiple topics. Do away with topic sections all together on single page and just tag them with their topics could be one way to approach it.

>I typically prefer spacing before a return.
Done

>There's a comma at the end of this file. Is the with state following continuation of this line? Or is it a separate block after?
Not sure how that got in there. Removed in latest commit.

>Should use () w/ print to be py3 compatible.
Done.

scripts/generate-kb-articles.py
```diff
@@ -0,0 +1,95 @@
...
+    convert_string = "### %s\n %s" % (article['title'], body)
+
+    return pypandoc.convert_text(convert_string, 'rst', 'md')
+
+
+def write_rst(kb_data):
+    """Write RST file to disk.
+    """
+    with open(RST_PATH, "w+") as rst_file:
+        rst_file.write(RST_HEADER)
```
m4dcoder: Please add a blank line between the with blocks. This makes it easier to read.

bigmstone: Done

m4dcoder:
> I can conceive of how that could be useful, but I am not the best person to answer that. It does beg the question of how to organize them on a page if they are a part of multiple topics. Do away with topic sections all together on single page and just tag them with their topics could be one way to approach it.

I think we can keep it simple for now w/ a single topic and revisit this later then.

> It's metadata from reamaze. In this case status of 0 means the article was in the published state.

So reamaze has published, draft, internal status.  I think instead of integer, we can make this status be enum of the actual text values and translate that to integer to reamaze in the script.

m4dcoder:
Also, what happens if the topic doesn't exist in reamaze?

bigmstone:
>Also, what happens if the topic doesn't exist in reamaze?

This logic will be handled by the workflow. The idea was that it will check if the article exists or not. If it doesn't make a call to create it and publish it. If it does exist just ignore it. I figured we could add more complexity later on if we wanted to also have it trigger to update articles, etc. but we'll need to figure out how to track the articles. I didn't see (and wasn't looking hard) in the API call to reamaze that there was any static unique identifier.

PR #284, opened 2016-09-30
Updating doc with working way of accessing attributes in webhook payload

sidkrishna:
PR for STORM-2409:
While trying out the generic webhook, it looks like that way an attribute in the payload needs to be accessed in the rule is different from what is given in the documentation.
As per doc, attribute being passed in the webhook as payload should be accessed as follows:
"trigger.body.payload.attribute1". This does not seem to work. If one uses ""trigger.attribute1" instead, it works. Updating doc to reflect the same. 
There is a possibility that ""trigger.body.payload.attribute1" is the correct expected method, and instead of doc, the ST2 code needs to be modified - will need comments/confirmation on this.

### DIFF
--- a/docs/source/webhooks.rst
+++ b/docs/source/webhooks.rst
@@ -147,7 +147,7 @@ Rule:
         type: "mypack.mytrigger"
 
     criteria:
-        trigger.body.payload.attribute1:
+        trigger.attribute1:
             type: "equals"
             pattern: "value1"
 


LindsayHill:
This now conflicts with https://github.com/StackStorm/st2docs/pull/285

Can you update this PR?

sidkrishna:
Updated the PR.

emedvedev:
LGTM, although I'd ask @Kami to take a look as well, just to make sure it's the intended way to do it and not a bug in the code.

LindsayHill:
Seems to be a lack of clarity here. Looks like `trigger.body.attribute1` might be OK (i.e. doesn't need `payload`). Probably need to do a bit more digging into how this is supposed to work

sidkrishna:
@Kami request you to pls have a look at this PR. Not sure whether it is actually an issue in the code itself.

LindsayHill:
@sidkrishna Kami is taking a break at the moment...when he gets back from his holidays he'll be able to confirm it.

Kami:
Thanks, I will look in the code later today and verify which approach is correct.

I know that in the past, you did need to use trigger.body.payload, but it's possible that something has changed or similar.

Kami:
Thanks for this - I have confirmed your change is correct.

I assume the example was incorrectly referring to a custom registered webhook where this notation is correct instead of generic st2 webhook which uses a different notation (no body).

Kami:
Merged. Will also cherry pick into v2.0.

Thanks again!

PR #285, opened 2016-09-30
Clarify the generic webhook purpose

emedvedev:
The distinction between generic and custom webhooks (or rather the purpose of a generic webhook) is a bit unclear to users, trying to clarify: moved custom webhooks to the top as they are easier to understand (and arguably they are used more widely), and rewrote the generic webhook description a bit.

P.S. Props to @eljas on the community channel, who was patient enough to go through this with me. :)

### DIFF
--- a/docs/source/webhooks.rst
+++ b/docs/source/webhooks.rst
@@ -59,50 +59,11 @@ is determined based on the value of the ``Content-Type`` header (``application/j
 All the examples below assume JSON and as such, provide ``application/json`` for the
 ``Content-Type`` header value.
 
-Using a Generic Webhook
------------------------
-
-By default, a generic webhook with the name ``st2`` is already registered. This
-webhook allows you to push arbitrary triggers to the API.
-
-The body of this request needs to be JSON and must contain the following attributes:
-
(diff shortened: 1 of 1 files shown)


docs/source/webhooks.rst
```diff
@@ -156,6 +117,47 @@ Rule:
         ref: "mypack.myaction"
         parameters:
     ...
+    
+Using a Generic Webhook
+-----------------------
+
+By default, a special-purpose webhook with the name ``st2`` is already registered. Instead
+of using ``st2.core.webhook``, it allows you to specify any trigger that is known to |st2| 
+(either by default or from a custom sensor), so you can use it to trigger rules that aren’t
```
Kami: Yeah, it's great to point this part out.

Maybe also a good idea to say custom sensor / trigger since since v1.6 or so we now also support triggers/ directory with trigger definitions.

Kami:
LGTM, +1

docs/source/webhooks.rst
emedvedev: Done :)

PR #289, opened 2016-10-11
RFR: Monitoring documentation

LindsayHill:
First cut of monitoring documentation. Addresses ST2 issue  https://github.com/StackStorm/st2/issues/2681

It's not fantastic, but it's a start. Want to get this merged, so we can at least start getting feedback on it.

Note that I've also made a few small changes to the syslog configuration docs.


docs/source/monitoring.rst
```diff
@@ -0,0 +1,153 @@
...
+    mistral-api PID: 1006
+    mistral-api PID: 1228
+    mistral-api PID: 1231
+    $
+
+In a distributed system, only some of these processes will be running on each system. In the example here ``st2chatops`` is not configured on this system.
+
+Tools such as nagios or check_mk can be used to monitor the process list. Some services spawn more than one process. The exact number will depend upon your system configuration - e.g. ``st2actionrunner`` will spawn additional processes on a multi-core system. 
+
+Additional processes to monitor:
```
codyaray: possibly out of scope here, but I added "logstash" (or rsyslog, or whatever log ingest tool) to our internal processes to monitor as well

codyaray:
Speaking of... I think my mongo queries above are wrong. I'm currently querying status from `action_execution_d_b` but hitting an issue right now that makes me think we should be querying from `live_action_d_b` instead. Maybe someone from StackStorm can tell us which is the "right" one? :)

Kami:
@codyaray Yeah, in this case live_action_d_b table should be queried / monitored (action execution db object is created at the end).

LindsayHill:
Thanks, pushed updated to use `live_action_d_b`

PR #291, opened 2016-10-13
To see all predefined actions: st2 action list

punalpatel:
removed --pack=core

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -647,7 +647,7 @@ To see all predefined actions:
 
 ::
 
-    st2 action list --pack=core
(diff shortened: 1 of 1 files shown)


emedvedev:
The section you're editing describes core actions only, so the `--pack=core` argument should stay. You could approach it a bit differently: replace "To see all predefined actions" with "To see all core actions" for clarification, and that would probably be enough to remove confusion. :)

Btw, `st2 action list` itself is described under "Managing and Running Actions," the very first section on that page, so we got it covered.

LindsayHill:
Closing this in favour of #311

PR #293, opened 2016-10-14
Adding support for pluggable runners.

bigmstone:


DoriftoShoes:
+1

scripts/generate-runner-parameters-documentation.py
```diff
@@ -21,12 +21,42 @@
...
+def _get_runners_dir():
+    runner_loader = RunnersLoader()
+    runner_dir = os.path.join(CURRENT_DIR, '../st2/contrib/runners')
+
+    runner_types = runner_loader.get_runners([runner_dir])
+
+    return runner_types
+
+
+def _get_runners():
```
Kami: In theory you could also move this method `get_runners_from_dir` (or similar) into bootstrap module and re-use it, but I'm also fine with only having it here for now.

Kami:
Nice :+1:

PR #295, opened 2016-10-17
Explain the result tuple for python actions

emedvedev:
There's been a question in Slack the other day about actions returning `False`, so I wanted to clarify this in the docs.

I'll wait for someone else to review before merging.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -533,9 +533,9 @@ standard output. As you can see, user-supplied action parameters are passed to
 the ``run`` method as keyword arguments.
 
 If the ``run`` method finishes without exceptions, the execution is successful,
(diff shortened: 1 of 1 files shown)


humblearner:
While you are at it, can you please fix this too: https://github.com/StackStorm/st2docs/pull/295/files#diff-157d9a0f26f16d59cb8468550a11f495R527 ?? :)

emedvedev:
@humblearner done :)

humblearner:
LGTM

Kami:
Much better :+1: 

Probably also a good idea to add small note that this functionality is only available in v1.6 and above, etc.

Kami:
Let's also cherry pick this into v2.x.

Kami:
I cherry picked it now so we should be fine.

PR #296, opened 2016-10-17
Note about pluggable runners and migration step

lakshmi-kannan:


docs/source/upgrade_notes.rst
```diff
@@ -17,6 +17,11 @@ Upgrade Notes
   ``{{system.foo}}`` and ``{{user.foo}}`` are still supported for backward compatibility but
   will be deprecated in subsequent releases.
 
+* Runners are now `pluggable`. With this version, we are piloting an ability to register
```
Kami: LGTM, but we should also add note on how to register core runners in this section as well, imo  (that's what we did in the past with other things).

Kami:
LGTM

PR #297, opened 2016-10-18
Add missing parameter to git config

iax7:
key http.proxy was missing in the command.


Kami:
Good catch - thanks!

Kami:
Also cherry picked this change into v2.0 branch.

PR #299, opened 2016-10-19
Add notes about decrypt_kv

lakshmi-kannan:


Kami:
LGTM.

As mentioned in st2 PR I still have some concerns with regards to RBAC but this should suffice for now.

PR #303, opened 2016-10-25
changes to config.schema

LindsayHill:
Removed some `config.yaml` references, replaced with `config.schema.yaml`.

Want to start moving all packs over, so need to stop users from creating `config.yaml` based on docs in the first place.


Kami:
LGTM, :+1:

PR #304, opened 2016-10-27
Update docs and upgrade notes for pack naming and ref changes

Kami:
Documentation for https://github.com/StackStorm/st2/pull/2972 and other related pack changes.

PR #305, opened 2016-11-05
Update get_by_name reference

LindsayHill:
Seems that it should be `client.keys.get_by_name(key)` not `client.keys.get_by_name(name=key)`. See https://github.com/StackStorm/st2contrib/blob/6de75209fa4f1ae6b3d2d77e4e529e4d0298d971/packs/st2/actions/kv_delete.py for a code example.

Users are getting tripped up by it - e.g. https://stackstorm-community.slack.com/archives/community/p1478302946001063

### DIFF
--- a/docs/source/datastore.rst
+++ b/docs/source/datastore.rst
@@ -136,7 +136,7 @@ Get individual key value pair or list all:
 ::
 
     >>> keys = client.keys.get_all()
-    >>> os_keystone_endpoint = client.keys.get_by_name(name='os_keystone_endpoint')
+    >>> os_keystone_endpoint = client.keys.get_by_name(os_keystone_endpoint)
     >>> os_keystone_endpoint.value
     u'http://localhost:5000/v2.0'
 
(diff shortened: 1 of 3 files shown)


docs/source/datastore.rst
```diff
@@ -136,23 +136,23 @@ Get individual key value pair or list all:
 ::
 
     >>> keys = client.keys.get_all()
-    >>> os_keystone_endpoint = client.keys.get_by_name(name='os_keystone_endpoint')
+    >>> os_keystone_endpoint = client.keys.get_by_name(os_keystone_endpoint)
```
Kami: Good catch.

> st2client/models/core.py
> 281:    def get_by_name(self, name_or_id, **kwargs):

In general I prefer to be explicit and use keyword arguments so another option is to explicitly use `name_or_id` here (although, yeah, most other code doesn't use a keyword argument).

Kami:
LGTM :+1: 

Hope merging my branch won't cause too many merge conflict issues on your side when merging it into master.

LindsayHill:
Thanks @kami - looks like your changes in #304 aligned nicely, so no conflicts!

Issue #306, opened 2016-11-07
Full path to st2sensorcontainer needed in sensors.rst doc.


lakshmi-kannan:
https://docs.stackstorm.com/sensors.html#debugging-a-sensor-from-a-pack

```
root@stackstorm:/opt/stackstorm# st2sensorcontainer
st2sensorcontainer: command not found
```

LindsayHill:
Closed in #307

PR #307, opened 2016-11-08
Added full path to st2sensorcontainer

LindsayHill:
Addresses https://github.com/StackStorm/st2docs/issues/306

### DIFF
--- a/docs/source/sensors.rst
+++ b/docs/source/sensors.rst
@@ -218,8 +218,8 @@ If there are errors in registration, fix the errors and re-register them using s
 
 
 Once you like your sensor, you can promote it to a pack (if required) by creating a pack in
-/opt/stackstorm/packs/${pack_name} and moving the sensor artifacts (YAML and Python) to
-/opt/stackstorm/packs/${pack_name}/sensors/. See :doc:`/reference/packs` for how to create a pack.
+`/opt/stackstorm/packs/${pack_name}` and moving the sensor artifacts (YAML and Python) to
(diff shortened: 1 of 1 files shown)


Kami:
IIRC, we symlink all the components to `/usr/bin` / `/usr/local/bin` so if it's not in $PATH it must be some other issue.

In general I'm fine with this change as long as it's consistent everywhere and as long as we try to figure out why it wasn't originally in $PATH.

LindsayHill:
Mmmm, that's an interesting point actually about $PATH. It clearly wasn't in that user's path, and I've seen other reports of something similar. I might have a quick look around some test systems before merging this.

armab:
Just checked, and there is no `st2sensorcontainer` in `/usr/local/bin`, here is the full list:
https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.links

So the PR looks good. Probably confused with `service st2sensorcontainer`.


If we need `st2sensorcontainer` in `$PATH` - we can add it, but not sure if that's needed, considering its "debugging" nature.

Kami:
@armab Yeah, I was wrong and you are right - we only symlink the tools but not the actual services.

In this case @LindsayHill change is good and we should do the same everywhere else. Another option is to also symlink the services.

What do other people think?

armab:
From the other side, `st2-rule-tester` is already there: https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.links#L4 which is for debugging purposes :)

Kami:
@armab true, but rule-tester is a script (tool) and not a long living service.

Anyways :)

armab:
@kami Ah right, got it :+1:

PR #315, opened 2016-11-21
should be object, not objects

pengyao:
https://github.com/StackStorm/st2/blob/v2.0.1/st2common/st2common/util/casts.py#L95


dzimine:
Thank you!

PR #320, opened 2016-12-01
Add Ubuntu Xenial

armab:
A couple of changes from https://github.com/StackStorm/st2-packages/pull/348/files for `Ubuntu Xenial` Installation instructions.

Glad there is no need for a separated page.

> Will open the same PR (cherry-pick) against `v2.1` once we merge this.


lakshmi-kannan:
We'll need this in v2.1 as well

PR #321, opened 2016-12-01
Fix commit sha for the cloudflare example

emedvedev:
Gitrefs changed after the clean transfer, so the cloudflare example fails; reflecting in the docs.

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -78,7 +78,7 @@ By default, the latest version of a pack will be installed, but you can specify 
 
 .. code-block:: bash
 
-    st2 pack install cloudflare=77ee04e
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
I'd just say SHA and not use any particular sha. Also, we should pick this up for 2.1

emedvedev:
We could change it to something obviously bogus like `abcabc1`, but then again, judging from our Slack token called "CHANGEME", people _will_ try to use it 😛 Better have a real SHA you can copy and paste.

lakshmi-kannan:
Heh. Don't forget to pick this for 2.1

emedvedev:
Yes, sir!

PR #332, opened 2016-12-04
st2-self-check: remove `bc` package dependency

armab:
See https://github.com/StackStorm/st2tests/pull/87 and https://github.com/StackStorm/st2cd/pull/231 which replaces `bc` -> `awk` that already used in the same workflow for calculation.

> Ignoring docs sometimes leads to improvements :smiley:


humblearner:
👍

PR #335, opened 2016-12-05
config.yaml deprecation notice

Kami:
The title says it all.


Kami:
Will cherry-pick into v2.1.

PR #338, opened 2016-12-06
Github link points to right branch.

dzimine:
This is a test change to see if the CI works OK.


Kami:
Is this still needed?

LindsayHill:
I'm not sure if we should make this change or not. The problem is that people will follow the link, make the docs change, and then we'll miss that it's just against a specific branch, and never merge to master, so their changes will be lost. I'd rather see people make changes against master.

PR #342, opened 2016-12-07
Document the difference between the number and integer JSON schema type

Kami:
We don't document it anywhere in our docs and it's not immediate obvious to the users.

https://github.com/StackStorm/st2/issues/3099


LindsayHill:
Thanks for adding that @Kami - the json schema docs that we refer to don't clarify it either. This is helpful for users.

PR #344, opened 2016-12-07
Minor edit in the upgrade documentation

sidkrishna:
1. The package upgrade needs to be done first to get the migration scripts so the scripts can be executed on after package upgrade. Changing the order of the steps to reflect the same. 

2. Just adding some clarity on how an existing community pack can be updated using the new pack management CLI. If someone is following the upgrade notes without prior knowledge of pack management CLI, this would help.

### DIFF
--- a/docs/source/install/upgrades.rst
+++ b/docs/source/install/upgrades.rst
@@ -28,10 +28,7 @@ Upgrade Procedure
 
    sudo st2ctl stop
 
-2. Run the migration script (if any). See section below for |st2|
-   version-specific migration scripts.
-
-3. Upgrade |st2| packages (``st2``, ``st2web``, ``st2chatops``, ``st2mistral``
+2. Upgrade |st2| packages (``st2``, ``st2web``, ``st2chatops``, ``st2mistral``
    and ``bwc-enterprise`` using distro specific tools.
 
   Ubuntu:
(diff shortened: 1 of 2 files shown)


emedvedev:
LGTM

docs/source/install/upgrades.rst
```diff
@@ -28,10 +28,7 @@ Upgrade Procedure
 
    sudo st2ctl stop
 
-2. Run the migration script (if any). See section below for |st2|
```
dzimine: Why running migration script is no longer needed/listed? I didn't try this so I don't know, but @Kami knows, so pls check with him and once he's ok I am ok.

sidkrishna: @dzimine  Thanks for looking. I have changed the sequence and havn't removed step for running migration script. Basically, the migration scripts can be run once the upgrade of packages (eg. st2) has been done. The doc currently asks user to run the migration script first and then do the package upgrade which does not work.

@Kami  Could you please have a look at this PR and let me know if the changes are OK or not.

armab:
The PR is legit, good catch :+1: 

Upgrade `st2` package first and then apply migration script.

docs/source/install/upgrades.rst
sidkrishna: @Kami could you please have a quick look? thanks

Kami:
Thanks - the changes look good to me.

Will merge it into master and also cherry-pick into v2.1.

Kami:
Ah, ignore my last statement - PR was opened against v2.1 so I will do it vice-versa (cherry pick it into master).

Anyway, thanks again and good catch :)

PR #345, opened 2016-12-08
Pin sphinx version

enykeev:


LindsayHill:
Looks like this is actually related to sphinx_rtd_theme - see https://github.com/snide/sphinx_rtd_theme/pull/346

PR #351, opened 2016-12-13
Adding TCP ports required in system requirements doc

sidkrishna:
Updating system requirements with the list of TCP ports needed. This is as per https://github.com/StackStorm/st2-packages/pull/387

### DIFF
--- a/docs/source/install/system_requirements.rst
+++ b/docs/source/install/system_requirements.rst
@@ -31,4 +31,4 @@ for the best experience while testing or deploying |st2|:
 |  * Recommended EC2: **t2.medium**    | * Recommended EC2: **m4.xlarge**  |
 +--------------------------------------+-----------------------------------+
 
-|st2| requires the following TCP ports to be open for successful installation and operation: nginx (80, 443), mongodb (27017), rabbitmq (4369, 5672, 25672), and st2 (9100-9102).
+By default, |st2| and related services use these ports: 80, 443, 27017, 4369, 5432, 5672, 25672, 9100, 9101 and 9102. If any other services are currently using these ports, |st2| may fail to install or run correctly.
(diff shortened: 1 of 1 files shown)


docs/source/install/system_requirements.rst
```diff
@@ -30,3 +30,5 @@ for the best experience while testing or deploying |st2|:
 |  * 10GB storage                      | * 40GB storage                    |
 |  * Recommended EC2: **t2.medium**    | * Recommended EC2: **m4.xlarge**  |
 +--------------------------------------+-----------------------------------+
+
+|st2| requires the following TCP ports to be open for successful installation and operation: nginx (80, 443), mongodb (27017), rabbitmq (4369, 5672, 25672), and st2 (9100-9102).
```
armab: > the following TCP ports ~to be open~

I think it's a bit confusing, because it can be understood as opening ports in firewall.
And we don't need to expose any ports to public net, except of web.

So IMO something like:
> |st2| uses the following TCP ports [...]

would be more appropriate.

LindsayHill: Agreed with @armab. Maybe we should phrase it as: "By default, |st2| and related services use these ports: x, y, z. If any other services are currently using these ports, |st2| may fail to install or run correctly."

docs/source/install/system_requirements.rst
```diff
@@ -30,3 +30,5 @@ for the best experience while testing or deploying |st2|:
 |  * 10GB storage                      | * 40GB storage                    |
 |  * Recommended EC2: **t2.medium**    | * Recommended EC2: **m4.xlarge**  |
 +--------------------------------------+-----------------------------------+
+
+By default, |st2| and related services use these ports: 80, 443, 27017, 4369, 5432, 5672, 25672, 9100, 9101 and 9102. If any other services are currently using these ports, |st2| may fail to install or run correctly.
```
armab: I liked "group by service" approach in previous version, ex: `rabbitmq (4369, 5672, 25672)`.
Gives better picture, comparing to just listing all the ports.

armab:
:+1:

sidkrishna:
@armab @LindsayHill Have grouped it back by service. Hopefully this should be fine for merging..

LindsayHill:
I'm happier with it grouped. LGTM, will merge

PR #354, opened 2016-12-17
Missing https://

LindsayHill:
Same as #353, applied to v2.1 branch

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -59,7 +59,7 @@ ready-made packs submitted and maintained by the StackStorm community. There are
 packs for most of the popular cloud providers and DevOps tools, as well as more
 peculiar integrations (hello, ``tesla`` and ``hue``).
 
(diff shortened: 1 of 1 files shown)


Kami:
Good catch - LGTM

Kami:
> ./scripts/clone-st2.sh
> Cloning branch v2.1.1 of st2
> Cloning into 'st2'...

This seems wrongs, this should be a branch, not a tag, looking.

Kami:
`version.txt` looks fine so no idea what is going on.

LindsayHill:
Docs building is a little weird for x.y.z versions. Docs only have x.y.
There is a workflow somewhere that strips .z.

Could be related

Kami:
@LindsayHill This shouldn't matter, it should use version from `version.txt` which is correctly set to `v2.1`, but it some how read `v2.1.1` from somewhere so it got me confused.

Kami:
Fix for the docs version is in https://github.com/StackStorm/st2docs/pull/355.

PR #355, opened 2016-12-19
Only use major.minor version component when checking out st2.

Kami:
This change was missing.

It looks like it was committed to v2.0, but not to master and v2.1 which is not causing build failure - https://github.com/StackStorm/st2docs/commit/ccab691f8a4cc1b4c5b344c0d391e4c9e7f2c346

Will also merge it into master.


Kami:
Also cherry picked it into master, build is green now.

PR #359, opened 2016-12-20
Updating requirements to 2GB

LindsayHill:
1GB is no longer enough. Need at least 2GB, otherwise good chance of getting out of memory failures during install/operation. 

CF https://github.com/StackStorm/st2vagrant/pull/17


humblearner:
lgtm

PR #362, opened 2017-01-04
Add notice to changes in retry policy for workflow actions

m4dcoder:
Retry policy no longer supports actions executed under a workflow and a notice is added to the appropriate doc section.


Kami:
:+1:

PR #365, opened 2017-01-12
fixing install docs - start services section

sidkrishna:
Start services commands come up in the docs twice - once is explicit mention and the other comes from common/start_services.rst. Using the later only for consistency.

### DIFF
--- a/docs/source/install/common/start_services.rst
+++ b/docs/source/install/common/start_services.rst
@@ -2,6 +2,6 @@
 
     sudo st2ctl start
 
-* Register sensors and actions ::
+* Register sensors, rules and actions ::
 
-    st2ctl reload
+    sudo st2ctl reload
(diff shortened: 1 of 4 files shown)


docs/source/install/common/start_services.rst
```diff
@@ -2,6 +2,6 @@
 
     sudo st2ctl start
 
-* Register sensors and actions ::
+* Register sensors, rules and actions ::
```
sidkrishna: @Kami thanks for pointing out. If I check the "sudo st2ctl reload" output on st2 2.1.0, it says:

2017-01-17 08:47:25,560 INFO [-] =========================================================
2017-01-17 08:47:25,560 INFO [-] ############## Registering rules ########################
2017-01-17 08:47:25,560 INFO [-] =========================================================
2017-01-17 08:47:26,122 INFO [-] Registered 17 rules.

Am I missing something else?

sidkrishna: @Kami could you have a look at this pls?

Kami: Sorry, you are right.

We updated `st2ctl reload` to also register rules by default (but not actual `st2-register-content` script which is called by st2ctl).

So, LGTM

Kami:
It has also been already merged so it's fine :)

Thanks again.

PR #366, opened 2017-01-21
update the copyright date

userlocalhost:
This patch update the copyright date from '2016' to 'current year' which is set automatically.

Currently, the year describing of copyright is old and it's specified statically.

By this patch, you don't need to update it for each year any more.

Thank you.

### DIFF
--- a/docs/source/info.py
+++ b/docs/source/info.py
@@ -5,10 +5,12 @@
 # change from one version to another in conf.py should be 
 # placed here, otherwise you WILL break the build.
 
+from datetime import datetime
+
 master_doc = 'index'
 
 project = u'StackStorm'
-copyright = u'2016, StackStorm'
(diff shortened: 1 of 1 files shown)


LindsayHill:
Thanks!

Kami:
Good catch - thanks :)

docs/source/info.py
```diff
@@ -5,10 +5,12 @@
 # change from one version to another in conf.py should be 
 # placed here, otherwise you WILL break the build.
 
+from datetime import datetime
+
 master_doc = 'index'
 
 project = u'StackStorm'
-copyright = u'2016, StackStorm'
+copyright = u'%s, StackStorm' % (datetime.now().strftime("%Y"))
```
Kami: Perhaps we should do `2014 - %s`.

userlocalhost: Thank you for your comment. I got it.
I'll make another PR to correct this point.

PR #369, opened 2017-01-25
Include Ansible Installation method in docs

armab:
We just released massive update for [`ansible-st2`](https://github.com/StackStorm/ansible-st2) `v0.6.0` https://github.com/StackStorm/ansible-st2/releases which introduces full platform support (4 platforms) and includes a lot of testing, bug-fixing and polishing.

With this update I'm confident about the quality, additionally changes are tested against [4 platforms in Travis](https://travis-ci.org/StackStorm/ansible-st2).

I think it worth to include Ansible as one of the installation methods now :100:

### DIFF
--- a/docs/source/install/ansible.rst
+++ b/docs/source/install/ansible.rst
@@ -1,0 +1,29 @@
+Ansible Playbooks
+=================
+
+Ansible playbooks to install |st2|.
+
+Allows you to deploy and further configure |st2| installation on local or remote machines with Ansible configuration management tool.
+Playbooks source code is available as GitHub repository `ansible-st2 <https://github.com/StackStorm/ansible-st2>`_.
+
+---------------------------
+
+Supported platforms
+---------------------------
+* Ubuntu Trusty (14.04)
+* Ubuntu Xenial (16.04)
+* RHEL6 / CentOS6
+* RHEL7 / CentOS7
+
+Quick Start
+---------------------------
+.. sourcecode:: bash
+
+    git clone https://github.com/StackStorm/ansible-st2.git
+    cd ansible-st2
+
+    ansible-playbook stackstorm.yml
+
+
+Please refer to https://github.com/StackStorm/ansible-st2 for updates and more detailed examples, descriptions and code.
+Additionally if you're familiar with Ansible, found a bug or would like to propose a feature, - your contributions are very welcome!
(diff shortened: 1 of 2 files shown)


docs/source/install/ansible.rst
```diff
@@ -0,0 +1,29 @@
...
+---------------------------
+.. sourcecode:: bash
+
+    git clone https://github.com/StackStorm/ansible-st2.git
+    cd ansible-st2
+
+    ansible-playbook stackstorm.yml
+
+
+Please refer to https://github.com/StackStorm/ansible-st2 for updates and more detailed examples, descriptions and code.
```
Mierdin: Should this be a link too? Not sure how RTD will render this

LindsayHill: Bare links get rendered as a link - see https://docs.stackstorm.com/mistral_yaql.html?highlight=yaql#testing-yaql-expressions for an example. In the RST it appears as http://yaqluator.com/., and it gets rendered as a link

LindsayHill:
I'm not sure about putting it as a top-level menu item yet. It needs to have references from some other sections, e.g. /install/index.html. 

We also need to clean up the orphan file here https://github.com/StackStorm/st2docs/blob/master/docs/source/install/puppet_chef_salt_ansible.rst

docs/source/install/index.rst
```diff
@@ -54,6 +54,7 @@ Once it completes successfully, you will see the following output:
     Ubuntu 14.04 / 16.04 <deb>
     RHEL 7 / CentOS 7 <rhel7>
     RHEL 6 / CentOS 6 <rhel6>
+    Ansible Playbooks <ansible>
```
armab: @LindsayHill Sub-menu under "Installation" Is close to previous structure, see: https://docs.stackstorm.com/1.3/install/index.html
I'm for keeping it visible to get more users trying + testing this method for us, what could be especially beneficial considering we're running Ansible for own infra.

armab: Alternative approach is to polish and revert entire 'puppet_chef_salt_ansible.rst' page what possibly can encourage contributions for other installation methods. In a meantime I left that page if we reconsider in some future time. WDYT?

LindsayHill: Right now, there's not really enough content here to justify having a dedicated Ansible page. Either we put this content in another page (e.g. refer to it from the main install page), or we overhaul the older doc, push Ansible as the 'preferred/supported' way of doing it, with notes that older Puppet/Chef stuff exists

armab: My initial idea was to keep it short as possible and refer to GitHub repo.

But I can bring enough content from the `ansible-st2` README, - it's not a problem. Including more detailed installation examples, requirements and configuration vars.

LindsayHill: Yes, I think that bringing in some content from the README would give it enough 'meat'

armab: Added more content like descriptions, details and examples.

PR #370, opened 2017-01-26
Fixed formatting

LindsayHill:
> & < can just be used directly in the RST. We had some HTML codes, which showed up in the rendered docs as &gt;, which was not the desired output.


Kami:
Good catch :+1: 

I assume this was copied in from the blog so it got messed up.

PR #371, opened 2017-01-27
Change instructions on how to obtain winexe

LindsayHill:
Addresses st2 issue https://github.com/StackStorm/st2/issues/3104

Ideally we would have pre-built packages available, but this is an important first step.

Doesn't seem to be any good sources of scripts for building winexe 1.1 deb files.


Kami:
LGTM :+1:

Kami:
I will also cherry pick this into v2.1.

PR #376, opened 2017-02-03
Add openssl-devel as requirement when installing from source on Fedora

### LINKED ISSUE #375: Add 'openssl-devel' as requirement for Fedora when running from source
I've tried to run stackstorm from source, on a CentOS 7.3 machine, but failed during 'make requirements'.

After enabling pip output, I saw that gcc fails, due to a missing openssl-devel package.

For Ubuntu libssl-dev is mentioned in the requirements, but for Fedora openssl-devel isn't.

peschmae:
Pull request to fix #375 

I've just added 'openssl-devel' to the packages mentioned in the "Environment Prerequisites"

For me this solved the issue, that I couldn't compile on a CentOS7 machine.

### DIFF
--- a/docs/source/development/sources.rst
+++ b/docs/source/development/sources.rst
@@ -28,7 +28,7 @@ Fedora
 
 ::
 
-    yum install python-pip python-virtualenv python-tox gcc-c++ git-all screen icu libicu libicu-devel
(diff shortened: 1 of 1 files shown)


armab:
Thanks! :+1:

PR #377, opened 2017-02-03
Fix formatting of ssh-keygen commands

peschmae:
In the current version, all commands to be executed to create a user and add a SSH key, when running from source, are displayed on a single line.

This pull requests, fixes the formatting to be inline with the other command blocks on the same page.


LindsayHill:
Thanks @peschmae. I keep an eye out for this sort of thing, but they still slip through sometimes

I've added one more similar small change to this PR too. Once tests complete I'll merge it

PR #379, opened 2017-02-03
Refer YAQL Standard Library, and add a pro tip.

dzimine:
Based on the question: 

> is there a way to convert a string into an array? I'm pulling information from excel and it's coming in as a string, but the parameter I want to pass the variables too is type array.


dzimine:
@LindsayHill I am ashamed. Thanks!

PR #380, opened 2017-02-04
Added references to 'st2 login' and 'st2 whoami' commands

Mierdin:
This PR adds an explanation of the new `st2 login` and `st2 whoami` commands that were introduced in https://github.com/StackStorm/st2/pull/3123

### DIFF
--- a/docs/source/reference/cli.rst
+++ b/docs/source/reference/cli.rst
@@ -111,7 +111,7 @@ to point to the new username.
 .. NOTE::
    As with many other ``st2`` commands, ``st2 login`` will not create the configuration file
    for you. Keep this in mind especially if you're leveraging the ``--config-file`` CLI option,
-   or similar
+   or similar.
 
 Note that you can still use the "old" method of supplying both username and password
 in the configuration file if you wish.
--- a/docs/source/reference/cli.rst
+++ b/docs/source/reference/cli.rst
@@ -79,12 +79,31 @@ by passing ``--skip-config`` flag to the CLI as shown below:
 Authentication and auth token caching
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you specify username and password as authentication credentials in the
-configuration file, the CLI will try to use those credentials to authenticate and
-retrieve an auth token.
+The best way to authenticate to StackStorm via the CLI is using the
+``st2 login`` command. Similar to ``st2 auth``, you must provide your
+username and password:
 
-This auth token is by default cached on the local filesystem (in the ``~/.st2/token``
+    st2 login st2admin --password Password1!
+
+However, instead of just caching the token, this command will also modify the
+CLI configuration to include the referenced username. This way, future commands
+will know which cached token to use for authentication (since tokens are cached
+using the ``token-<username>`` format).
+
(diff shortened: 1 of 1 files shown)


docs/source/reference/cli.rst
```diff
@@ -79,13 +79,32 @@ by passing ``--skip-config`` flag to the CLI as shown below:
...
+using the ``token-<username>`` format).
+
+These auth tokens are by default cached on the local filesystem (in the ``~/.st2/token-<username>``
 file) and re-used for subsequent requests to the API service.
 
+You can also use the ``st2 whoami`` command for a quick look at who is the currently
+configured user.
+
+The ``st2 login`` command is a good option if you want an easy way to authenticate
+via the CLI without storing your password in plain text. Switching between users is also
```
dzimine: consider a ref to the CLI 101 section that talks about `plain text` option.

dzimine: and m.b. warn that it'll override the ./st2/config credentials?

Mierdin: Okay, I took a stab at adding these in https://github.com/StackStorm/st2docs/pull/380/commits/a6f461208856a10bf063f39c5051ba33f85ada88 - hopefully that's on target

LindsayHill:
LGTM, now that the crisis of the missing period has been resolved.

docs/source/reference/cli.rst
```diff
@@ -79,13 +79,43 @@ by passing ``--skip-config`` flag to the CLI as shown below:
 Authentication and auth token caching
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you specify username and password as authentication credentials in the
-configuration file, the CLI will try to use those credentials to authenticate and
-retrieve an auth token.
+If you don't wish to store your password in plain-text as shown in the previous section,
+the ``st2 login`` command offers an alternative. Similar to ``st2 auth``, you must provide your
+username and password:
```
Kami: It's maybe also worth adding a note that since this command doesn't store password by default in the config and it relies on a cached auth token it will only work until the token expires (which is 24 hours by default, IIRC).

After that, the user will need to re-login. And / or perhaps also document the `--write-password` option.

Kami: Besides that, LGTM.

docs/source/reference/cli.rst
```diff
@@ -79,13 +79,43 @@ by passing ``--skip-config`` flag to the CLI as shown below:
 Authentication and auth token caching
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you specify username and password as authentication credentials in the
-configuration file, the CLI will try to use those credentials to authenticate and
-retrieve an auth token.
+If you don't wish to store your password in plain-text as shown in the previous section,
```
Kami: Also, I think we should probably still keep part of this paragraph somewhere here explaining how authentication works in the CLI if username and password are specific in the config.

Mierdin: I moved this further down to line 116. Clarified in 8e3dc46

docs/source/reference/cli.rst
Mierdin: Good idea - added in 8e3dc46

PR #382, opened 2017-02-06
Re-generate available permission types table

Kami:
New permission type was introduced in https://github.com/StackStorm/st2/pull/3124.


LindsayHill:
LGTM

Issue #352, opened 2016-12-14
Issue with, "st2 pack" commands through proxy


mchenetz:
I had issues initially getting the, `st2 pack` commands working with a proxy. I had followed the instructions from:
https://docs.stackstorm.com/reference/proxy.html?highlight=proxy

When i ran a, `st2 pack search monitoring` I got the following output:
`
400 Client Error: Bad Request
MESSAGE: No results from the index: tried https://index.stackstorm.org/v1/index.json.
Status: [
   {
       "url": "https://index.stackstorm.org/v1/index.json", 
       "message": "SSLError(SSLError(SSLError(\"bad handshake: SysCallError(104, 'ECONNRESET')\",),),)", 
       "error": "unresponsive", 
       "packs": 0
   }
] for url: http://127.0.0.1:9101/v1/packs/index/search`

# Fix
add: HTTP_PROXY and HTTPS_PROXY to the environment
`HTTP_PROXY=http://username:password@proxy:port/`
`HTTPS_PROXY=http://username:password@proxy:port/`
`no_proxy="127.0.0.1"`

### RUN:
`sudo st2ctl  restart`

## Might be needed - I made these changes too:
Modify: /opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/packs.py

### ADD:
`proxies = {
  'http': 'http://username:password@proxy:port',
  'https': 'http://username:password@proxy:port',
}'

### Modify
`request = requests.get(index_url)`
### To
`request = requests.get(index_url, proxies=proxies)`

### RUN:
`sudo st2ctl  restart`

LindsayHill:
@sidkrishna could you take a look at this one when you get a chance? I'm not sure if the code changes are needed, or just the docs about setting environment variables.

conor-owens:
I saw this issue too.  Centos7 and Stackstorm 2.1.0. More info: https://gist.github.com/conor-owens/21d2b291251c9f89ecef579de9fb5221

Seems the various environment variables are not preserved when running the packs.py so the requests doesn't find or use the HTTPS_PROXY environment variable.

conor-owens:
I tried adding this to the packs.py to read the contents of /etc/environment:  
(ref: http://stackoverflow.com/questions/5971312/how-to-set-environment-variables-in-python)

`import re

LOG = logging.getLogger(__name__)

def _set_env_from_etc_environment(envFilePath):
    print('--packs.py:setEnvFromEnvFile --1-----------------------------------------------.')
[...]

PR #388, opened 2017-02-08
Added note for temp fix that allows pack install to work over proxies

Mierdin:
Should address both https://github.com/StackStorm/st2/issues/3137 and https://github.com/StackStorm/st2docs/issues/352 well enough for now. Following these instructions is the best way to get `st2 pack install` working without modifying actual Python code.

A more permanent solution to this problem is being considered for a future release.


Kami:
LGTM.

Although yeah, a bit nasty needing to edit code on disk so we should look into a proper fix as soon as v2.2.0 is out.

PR #391, opened 2017-02-14
Enable wrapping for table cells

LindsayHill:
@enykeev what do you think of this? 

This change over-rides the `nowrap` behavior on table cell content. 

Before the change, we have a horizontal scrollbar, like this: (from docs.stackstorm.com/mistral.html)
table_nowrap

Afterwards, it looks like this:

table_wrap

I prefer the second version. From an initial look at our other tables, this change generally looks better. I haven't found any counter-examples yet.

I also have another reason for wanting this - I want to auto-generate some docs based upon action definitions, and it will use tables. If I can get the content to auto-wrap, it will be much simpler to generate my tables.

Couple of things:
1.  Min-width may not be right here - you might prefer a different value

2. This is one way to override the default CSS, but you might prefer a different approach. I have no opinion on the 'right' way to do it.

### DIFF
--- a/docs/source/_static/theme_overrides.css
+++ b/docs/source/_static/theme_overrides.css
@@ -1,0 +1,13 @@
+/* override table width restrictions */
+@media screen and (min-width: 767px) {
+
+   .wy-table-responsive table td {
+      /* !important prevents the common CSS stylesheets from overriding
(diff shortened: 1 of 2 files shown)


armab:
:+1: 
Some big tables with horizontal scrollbar looked really ugly.

PR #392, opened 2017-02-16
Update upgrade section to clarify Mistral database upgrade

m4dcoder:
If mistral services are restarted before the mistral-db-manage commands are run, the mistral database upgrade will fail. Update appropriate section to highlight precaution.

### DIFF
--- a/docs/source/install/upgrades.rst
+++ b/docs/source/install/upgrades.rst
@@ -43,15 +43,21 @@ Upgrade Procedure
 
      sudo yum update $PKG_NAME
 
-3. Run the migration script (if any). See section below for |st2| 
-   version-specific migration scripts.
-
-4. Upgrade Mistral database.
+3. Upgrade Mistral database.
 
 .. sourcecode:: bash
 
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
+
+.. note::
+
+    The mistral and mistral-api services must be stopped at time of upgrade. If the services are
+    restarted before the mistral-db-manage commands are run, then the
+    ``mistral-db-manage upgrade head`` command may fail.
+
+4. Run the migration script (if any). See section below for |st2|
+   version-specific migration scripts.
 
 5. Start |st2| services.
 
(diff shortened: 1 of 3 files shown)


docs/source/install/upgrades.rst
```diff
@@ -82,6 +88,34 @@ v2.2
   executions_v2, either use psql or install an older version of the python-mistralclient in a
   separate python virtual environment.
 
+.. note::
+
+    Please be sure to follow the general steps listed above to do the database upgrade.
+
+*  If you're seeing an error ``event_triggers_v2 already exists`` when running
+   ``mistral-db-manage upgrade head``, this means the mistral services started before the
```
armab: At a high-level this behavior is horrible, so much error-prone and unfriendly in terms of package upgrade story :/

docs/source/install/upgrades.rst
```diff
@@ -43,16 +43,22 @@ Upgrade Procedure
...
-3. Run the migration script (if any). See section below for |st2| 
-   version-specific migration scripts.
-
-4. Upgrade Mistral database.
+3. Upgrade Mistral database.
 
 .. sourcecode:: bash
 
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head
   /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate
```
armab: I guess these 2 DB upgrade commands should be run on every new st2mistral version (eg the script won't fail if DB is already latest and it will upgrade if DB upgrade is really needed).

What if we include this logic in `st2mistral` packaging?

```sh
# best effort
# don't hang entire package upgrade with error if some of the commands failed (ex: postgresql is not yet available)
on every `st2mistral` package upgrade:
  stop mistral services
  mistral-db-manage upgrade head
  mistral-db-manage populate
  start mistral services
```

Any corner cases? WDYT?

docs/source/install/upgrades.rst
```diff
@@ -82,6 +88,34 @@ v2.2
   executions_v2, either use psql or install an older version of the python-mistralclient in a
   separate python virtual environment.
 
+.. note::
```
lakshmi-kannan: Maybe we should use .. warning:: flag instead so this shows in a red box?

docs/source/install/upgrades.rst
lakshmi-kannan: Yeah, I think we will all agree with you :).

docs/source/install/upgrades.rst
lakshmi-kannan: I think this is a long due change. If we agree these steps should be run on every install (not just upgrade), we can just put these in post-inst (forgot the rpm equivalent).

Kami:
> If we agree these steps should be run on every install (not just upgrade), we can just put these in post-inst (forgot the rpm equivalent).

Agreed (and AFAIK this step is idempotent so it should be fine as well).

docs/source/install/upgrades.rst
m4dcoder: The command will require the /etc/mistral/mistral.conf to exist.

m4dcoder:
Warning flag added

PR #393, opened 2017-02-16
Add some additional docs on running pack tests

Kami:
This pull request documents the new ``-j`` flag and clarifies how to set `ST2_REPO_PATH` environment variable to point to st2 git repo checkout in case StackStorm components are not already in ``PYTHONPATH`` (e.g. when running on a system where StackStorm is not installed).


lakshmi-kannan:
Is -j for --jolly? 

LGTM

Kami:
-j as in jock :P

PR #395, opened 2017-02-16
Adding section to describe pack_context usage

bigmstone:
Refers to  StackStorm/st2#3183


LindsayHill:
Ah, much better - checks passing now

PR #401, opened 2017-02-22
Document that the installer script now generates random passwords for MongoDB and PostgreSQL

Kami:
The title says it all.

That's the upgrade notes part of it, for v2.2.1, I think we should also do what I described here - https://github.com/StackStorm/discussions/issues/225#issuecomment-281602924 (https://github.com/StackStorm/st2/issues/3203).


LindsayHill:
LGTM - guess we don't want to merge until we're publishing 2.2?

Kami:
Yeah, the headline change - I can remove it and we can then do it in v2.2 branch, I'm fine either way.

Kami:
Some how forgot to merge this before v2.2.0 so I will also cherry-pick it into v2.2 branch.

PR #404, opened 2017-02-27
WIP: Autogenerate Network Essentials docs

LindsayHill:
WIP script for auto-generating network essentials docs from metadata.

Note that this script is not run with st2docs builds, only bwcdocs.


LindsayHill:
Checks will fail until network essentials is moved to ST2 Exchange & made public, and script is modified to use new location

LindsayHill:
For now, don't want to merge this into regular Makefile

PR #412, opened 2017-03-03
Add more example for YAQL expressions

dpasupathi:
Use context and pass variables to the yaql expression, useful in defining yaql expression in the same task.


Kami:
Merged, thanks!

PR #413, opened 2017-03-04
Added Datastore Crypto Key instructions to installation

LindsayHill:
We didn't cover datastore crypto key generation in the manual installation section, so it could have been missed by people doing manual setups.


Kami:
LGTM :+1:

PR #415, opened 2017-03-05
Fixed typo with parameter vs payload

LindsayHill:
Same as #414, but for master. We referred to system.validate_trigger_parameters and parameters_schema twice, rather than system.validate_trigger_payload and payload_schema


Kami:
Good catch :+1:

PR #416, opened 2017-03-06
Document which coordination backends we now support

Kami:
We now ship latest version of `tooz` in StackStorm master and as such, more coordination backends are supported.

Part of https://github.com/StackStorm/st2/pull/3269.


LindsayHill:
LGTM, with one small language change made.

Issue #390, opened 2017-02-13
Use quotes everywhere where passwords are used/generated


LindsayHill:
Documentation should make it clear that passwords, etc should be quoted when used/generated via CLI.

This is to avoid issues like this https://github.com/StackStorm/st2/issues/3216 where the intention is to have a password like `pas$word`.

If a user enters `st2 auth user -p pas$word`, the shell will expand $word, usually to a blank string. It should be entered as `st2 auth user -p 'pas$word'`. Similarly when `htpasswd` is used to generate a password hash.

Locations such as this should include a note on escaping passwords https://docs.stackstorm.com/authentication.html?highlight=st2%20auth#usage, and we should always use single quotes around passwords where we use `st2 auth` in the docs.

PR #417, opened 2017-03-09
Wrap passwords in single quotes

LindsayHill:
Some users have been confused when they set passwords containing special chars such as `$`, as they don't seem to work using the Web UI. This is due to the shell interpreting some characters.

This change wraps all passwords in single quotes, and adds a note about it.

Addresses #390

### DIFF
--- a/docs/source/__auth_usage.rst
+++ b/docs/source/__auth_usage.rst
@@ -9,6 +9,11 @@ response. ::
     st2 auth yourusename
     Password:
 
+.. note::
+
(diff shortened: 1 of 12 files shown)


Kami:
LGTM, but it's also worth noting this would still cause issues if `'` character is used for password, but should probably be fine for now :D

LindsayHill:
@Kami yeah, I thought about that...but figured it was better to at least do something now to highlight to people that special chars need handling, since we had a couple of user reports of issues.

Didn't want to go too deep down the rabbit hole of shell escaping

PR #423, opened 2017-03-21
Update st2chatops role in ansible based installation

humblearner:


LindsayHill:
LGTM

docs/source/install/ansible.rst
```diff
@@ -88,6 +91,14 @@ Below is more advanced example to customize |st2| deployment:
         - name: Install st2web
           role: st2web
 
+        - name: Install st2chatops with "slack" hubot adapter
+          role: st2chatops
+          vars:
+            st2chatops_version: latest
+            st2chatops_st2_api_key: CHANGE-ME-PLEASE # This can be generated using "st2 apikey create -k"
+            st2chatops_hubot_adapter: slack
+            st2chatops_config: {"HUBOT_SLACK_TOKEN":"xoxb-CHANGE-ME-PLEASE"}
```
armab: Please use consistent yaml notation instead of json:
```yaml
vars:
  st2chatops_config:
    HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
```

armab: Overall looks good to me, after changing it ^^.

humblearner: @armab: ^^ This will not work:

This is not json, this is the hash used here: https://github.com/StackStorm/ansible-st2/blob/master/roles/st2chatops/defaults/main.yml#L17. Ansible considers it as a key value pair instead of hash:
```yaml
HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
```

armab: Can you double check?

I guess there are 2 complex data types in yaml: Sequences and Mappings. And for Mappings we can call it whatever: "Hash", "Dictionary" or key-value, - everything is the same in yaml. So I think it doesn't matters.

See: http://docs.ansible.com/ansible/YAMLSyntax.html#yaml-basics

For reference, I'm using it here: https://github.com/StackStorm/ansible-st2/pull/121/files#diff-7df2362a185e65ecc0b7dcd23cb3fd4cR11
with the following yaml example:
```yaml
vars:
  # https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample
  st2_config:
    auth:
      debug: True
      enable: True
    messaging:
      url: amqp://st2:st2@127.0.0.1:5672//
```

humblearner: You are right, the assertion in the play was not correct: https://github.com/StackStorm/ansible-st2/pull/129/commits/2cabe7ee42be6570014448881a2853fd4fe9cb66#diff-b9eaf1c4002cf73f79f7506ed1c4f718L13.

PR #431, opened 2017-03-24
ST2Flow Placeholders

LindsayHill:
Prep change for Flow Docs.

This puts placeholders in the ST2 docs, so that when https://github.com/StackStorm/ipfabric-docs/pull/64 gets merged, Flow docs will show up in bwc-docs.brocade.com, not docs.stackstorm.com

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -24,6 +24,10 @@ Contents:
     datastore
     chatops/index
 
+.. include:: _includes/flow.rst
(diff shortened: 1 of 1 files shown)


LindsayHill:
Also note: This moves the "Automation Suites" section above the "Advanced Topics" section. Again, this only applies to bwc-docs.brocade.com, not docs.stackstorm.com.

PR #432, opened 2017-03-24
Updated roadmap

LindsayHill:
Fixed formatting issue with nested lists, and added notes for v2.2


LindsayHill:
I'd almost forgotten about it...

Issue #368, opened 2017-01-25
Installation docs should stress requirement for CLEAN system for one-line install


LindsayHill:
The installation docs https://docs.stackstorm.com/install/ say "Get yourself a clean 64-bit Linux box that fits the system requirements."

For a one-line install, it assumes a clean, basic system. But people try to run the installer on customized systems that have things like non-standard authentication and $HOME setups, and this breaks. It is possible to install ST2 on those systems, but it needs to be a manual install, not a one-line install. 

We have added checks to the one-line install script to look for things like disk space, ports in use, etc. That helps, but we can't catch everything. 

We need to update the docs to stress this more clearly - a one-line install only works on an unadulterated Linux system. Perhaps a warning box, or similar?

LindsayHill:
Resolved with #433

PR #433, opened 2017-03-25
Stress requirements for clean system

LindsayHill:
More strongly stress the requirement for a **clean** system for using the one-line install.

Fixed a couple of minor language things. 

Addresses https://github.com/StackStorm/st2docs/issues/368

/cc @emedvedev

### DIFF
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -1,7 +1,7 @@
 Installation
 ============
 
-That's OK! You're busy, we get it. How do you just get started? Get yourself a clean 64-bit Linux
+That's OK! You're busy, we get it. How do you just get started? Get yourself a **clean** 64-bit Linux
 box that fits the :doc:`system requirements <system_requirements>`. Make sure that ``curl`` is up to date
 using ``sudo apt-get install curl`` on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run
 this command:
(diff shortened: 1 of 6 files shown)


docs/source/install/index.rst
```diff
@@ -10,10 +10,21 @@ this command:
 
    curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password='Ch@ngeMe'
 
+.. danger::
+
+  The installation script is an opinionated installation of |st2|. It assumes that you have a clean, basic
+  installation of Ubuntu or RHEL/CentOS, similar to what you get with a basic installation from ISO. If
+  you are trying to install |st2| on a server with many existing services, or maybe a 'pre-loved' VM that
```
emedvedev: > If you are trying to install ... that you will run into problems.

Looks like there's a typo or an unfinished edit here.

emedvedev:
:heart:

docs/source/install/index.rst
LindsayHill: Whoops, there is indeed. Mobile now, will fix it tomorrow

docs/source/install/index.rst
```diff
@@ -1,7 +1,7 @@
 Installation
 ============
 
-That's OK! You're busy, we get it. How do you just get started? Get yourself a clean 64-bit Linux
+That's OK! You're busy, we get it. How do you just get started? Get yourself a **clean** 64-bit Linux
```
armab: !!! :)

Kami:
LGTM :+1: 

Maybe also not a bad idea to explicitly point out that if for some reason installer script fails half way through, user should re-run it on a new clean server.

LindsayHill:
Good point @Kami will add another update about that

PR #435, opened 2017-03-27
Note about script not being idempotent

LindsayHill:
Added note about script not being idempotent, per https://github.com/StackStorm/st2docs/pull/433#issuecomment-289275595


Kami:
:+1:

PR #436, opened 2017-03-28
updated dependencies; fixed search

LindsayHill:
We pinned sphinx back in https://github.com/StackStorm/st2docs/pull/345, due to broken search (https://github.com/rtfd/sphinx_rtd_theme/pull/346). 

This is fixed with changes to `layout.html` and `layout_old.html`. With these changes we don't need to pin our sphinx version anymore, and can get new features & bug fixes again.

I've also changed the sphinx-autobuild version. We were using our own fork, which pinned the livereload version. That fork is now very out of date. I would rather use the upstream version unless we still have a specific need to use the old one. That doesn't seem to apply anymore. 

With this combination of new sphinx versions & theme changes, I can search again when viewing local live docs.

I've also made a small change to `conf.py` to remove this message when building docs:
```
| loading intersphinx inventory from http://docs.python.org/objects.inv...
| intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/2/objects.inv
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-sphinx
+sphinx>=1.5.3,<1.6
 sphinx-autobuild
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-sphinx==1.4.9
-git+https://github.com/StackStorm/sphinx-autobuild.git@develop
+sphinx
+sphinx-autobuild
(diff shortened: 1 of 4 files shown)


Kami:
Cool, LGTM :+1: 

I assume you verified that docs built with latest version look and work fine? :)

requirements.txt
```diff
@@ -1,2 +1,2 @@
-sphinx==1.4.9
-git+https://github.com/StackStorm/sphinx-autobuild.git@develop
+sphinx
```
Kami: Personally I would still a bit better if we at least pin it to the latest major version - `sphinx>=1.5.3,<1.6`.

This way we can avoid surprises if a new version pushes a backward incompatible or breaking change (we had this happened many times in st2 when we didn't pin dependencies to specific version and we wasted a lot of times debugging such issues).

LindsayHill: Yeah, that's a good point. I'll make that change. I've also done some testing and it looks good, but I want to do a little more testing here on a 'clean' system to verify it, then we can merge it.

PR #437, opened 2017-03-29
upgrade pip version;fix make distclean

LindsayHill:
We upgraded st2 to use newer pip version last August. But we didn't update the st2docs/Makefile. This change will make st2docs use the same pip version. 

I also fixed the `make distclean` target. Previously it would fail with errors like this:
```
(virtualenv) st2docs lhill$ make distclean
make: *** No rule to make target `clean', needed by `distclean'.  Stop.
```


Kami:
LGTM :+1:

Issue #438, opened 2017-03-29
Add package dependencies for st2docs build and run


sidkrishna:
The following packages are required for successfully build and run st2docs locally on linux:
`sudo apt-get install libpython-dev libssl-dev`

The README needs to be updated accordingly.

PR #440, opened 2017-03-30
Adding package dependencies for st2docs in README

sidkrishna:
Fixing issue https://github.com/StackStorm/st2docs/issues/438

The mentioned packages are required on the linux host where st2docs are being built and run. Without these, the build fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,10 @@ Product documentation for StackStorm is maintained in this repository. These doc
 Follows these steps to build the docs locally:
 
 Install the dependencies:
+
 For Debian/Ubuntu: ``sudo apt-get install libpython-dev libssl-dev``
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,6 +26,12 @@ Product documentation for StackStorm is maintained in this repository. These doc
 #### Build locally on Linux
 Follows these steps to build the docs locally:
 
+Install the dependencies:
+
+For Debian/Ubuntu: ``sudo apt-get install libpython-dev libssl-dev``
+
+For RHEL/CentOS: `` sudo yum install libpython-dev libssl-dev``
```
LindsayHill: Package names are different on RHEL/CentOS. You'll need to check, but I think it's python-devel and openssl-devel.

sidkrishna:
@LindsayHill Tested on CentOS and corrected required packages. Pls have a look.

PR #442, opened 2017-03-30
Explaining use case for execution API instead of webhooks

### LINKED ISSUE #441: Explain use case for directly using execution API
Every few weeks, someone asks about a specific use case, and it would be nice to have a doc that explicitly mentions this.

Basically the "traditional" flow of triggering a rule via webhook, and in turn, calling an action to do some work, is a bit too asynchronous for some - primarily because a webhook doesn't guarantee an execution ID. There are times when users just want to execute an action and get a result back.

When this comes up in slack, we usually just direct the user to the execution API to run an action directly. For instance, you can run `st2 --debug run mypack.myaction` and see how it executes the action, and periodically queries for the status of the execution.

So we should add something to docs that explains this, so we can point users to it instead of explain. It might also be useful to include a few points about the nginx configuration, as it might be believed that webhooks are the only external service exposed to the network (see [this thread](https://stackstorm-community.slack.com/archives/C066APT88/p1490908680789186))

Mierdin:
We get questions in support often about how to use webhooks to get information out of stackstorm. This is a bit of a misnomer, because webhooks are intentionally unidirectional - usually what the user wants is to simply run an action directly. So this PR adds a section to the docs that explains that the executions API is what they're looking for

This is still a WIP because I'm pretty unsure about the wording, as well as the location of this addition. Seeking feedback on both and more.

Closes #441

### DIFF
--- a/docs/source/webhooks.rst
+++ b/docs/source/webhooks.rst
@@ -173,22 +173,62 @@ When Not to Use Webhooks
 
 While webhooks are useful, they do have two drawbacks:
 
-* **Not Bidirectional**  - Webhooks simply send data into stackstorm. So if you want data back from   StackStorm, or an action, you'll have to get that data in an asynchronous fashion
-* **No Guarantee of Execution** - Webhooks in StackStorm do not guarantee an execution. This all depends on configuration of the rule that registered the webhook.
+* **Not Bidirectional**  - Webhooks simply submit data into |st2|. So if you want data back from |st2|,
+  or an action execution ID, you'll have to get that data in an asynchronous fashion.
+* **No Guarantee of Execution** - Webhooks in StackStorm do not guarantee an execution. It depends on
+  the configuration of the rule that registered the webhook - based upon the webhook contents, it may
+  not execute any action, or may execute multiple actions.
 
-If you're looking for guaranteed responses, it might be better to directly leverage the
-execution API to directly run actions and query for a response, in the same way the ``st2 run``
-command does behind the scenes. We can get a little insight into this using the ``--debug`` flag:
+If you always want to execute a specific action or workflow, and/or you're looking for guaranteed response,
+you could leverage the ``/v1/executions`` API. This is the same as running an action from the CLI with
+``st2 run <mypack>.<myaction>``. We can get a little insight into this using the ``--debug`` flag:
 
 .. sourcecode:: bash
 
     st2 --debug run core.local "date"
+    2017-03-31 08:21:18,706  DEBUG - Using cached token from file "/home/ubuntu/.st2/token-st2admin"
+    # -------- begin 140183979680208 request ----------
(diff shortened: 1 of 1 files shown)


docs/source/webhooks.rst
```diff
@@ -173,22 +173,62 @@ When Not to Use Webhooks
...
-execution API to directly run actions and query for a response, in the same way the ``st2 run``
-command does behind the scenes. We can get a little insight into this using the ``--debug`` flag:
+If you always want to execute a specific action or workflow, and/or you're looking for guaranteed response,
+you could leverage the ``/v1/executions`` API. This is the same as running an action from the CLI with
+``st2 run <mypack>.<myaction>``. We can get a little insight into this using the ``--debug`` flag:
 
 .. sourcecode:: bash
 
     st2 --debug run core.local "date"
+    2017-03-31 08:21:18,706  DEBUG - Using cached token from file "/home/ubuntu/.st2/token-st2admin"
```
Mierdin: I was going to stay away from this because of the length, but after seeing how this is rendered, it's not so bad. If users really want to see this output they can scroll to the right. Or actually run the commands for themselves, which is visible in the text window.

docs/source/webhooks.rst
```diff
@@ -173,22 +173,62 @@ When Not to Use Webhooks
...
+  or an action execution ID, you'll have to get that data in an asynchronous fashion.
+* **No Guarantee of Execution** - Webhooks in StackStorm do not guarantee an execution. It depends on
+  the configuration of the rule that registered the webhook - based upon the webhook contents, it may
+  not execute any action, or may execute multiple actions.
 
-If you're looking for guaranteed responses, it might be better to directly leverage the
-execution API to directly run actions and query for a response, in the same way the ``st2 run``
-command does behind the scenes. We can get a little insight into this using the ``--debug`` flag:
+If you always want to execute a specific action or workflow, and/or you're looking for guaranteed response,
+you could leverage the ``/v1/executions`` API. This is the same as running an action from the CLI with
```
Mierdin: Just a nitpick, really - do we want to say `v1` here? I'm not super plugged in to the upcoming API changes in 2.3 but I assume this will have a new version at some point in the future, so was just planning to refer to it as the "executions API" so we don't have to revisit this. The output below will also contain the correct URL for their reference

docs/source/webhooks.rst
LindsayHill: Yeah, the scrolling can be a problem. That's why I changed the last curl call around a bit, to move the URI to nearer the front, to make it more obvious.

docs/source/webhooks.rst
LindsayHill: I could go either way. All our examples will need to be updated when we release the new API version.

Mierdin: Ah, in that case, this is fine.

Mierdin:
@LindsayHill I'm 👍 on this if you are.

Issue #403, opened 2017-02-27
Control search engine indexing


Mierdin:
Googling for stackstorm concepts very often leads to very old versions of our docs. We should take steps to ensure that only the latest version comes up in search queries

armab:
:+1:
Something to discuss with @alexmakarski

LindsayHill:
Agreed. Need to have banner across top of old docs saying "THESE ARE OLD" or something like that - see the way http://docs.influxdata.com/influxdb/v1.1/ does it.

Probably needs `robots.txt` or similar updated to stop google indexing old versions?

Mierdin:
Personally I would prefer to make it difficult to even land on an old page. So I'd do both - put up a big banner saying "this is old", similar to how influx does it, but in addition, force the user to go out of their way to get to an old doc. So they won't even get there unless they're meaningfully doing it.

I mean right now, the version isn't even part of the URL - I haven't had time to dig into why this is, but our docs will [display an old version of the docs even if the URL says "latest"](http://imgur.com/a/oN4EY). I got to this page simply by googling "stackstorm actions". That definitely shouldn't happen

alexmakarski:
@LindsayHill I wouldn't hastily de-index the old stuff, especially if it's bringing in traffic. I think in Influx approach is more sound.

LindsayHill:
@Mierdin I think that you got those old versions because of some page re-arrangement we did. What was /runners.html got moved to /reference/runners.html. It looks like the build processes are only additive though - they're not deleting previously built content.

Mierdin:
If the old stuff is bringing in traffic, is that really traffic that we want? At best, those situations are providing two google results - one with the correct version, and one with the old version - forcing the user to choose. At worst, users are finding references to concepts that have either radically changed or simply don't exist anymore. I don't think that traffic is what we're after.

@LindsayHill is probably right, and we just need to clean up the old stuff somehow

LindsayHill:
I don't know if the cleanup is something easily automated (e.g. do a `make clean; make docs`). Depends how the sync with S3 step is done I guess - I don't know the details of that workflow

armab:
`robots.txt`:
```txt
User-agent: *
Disallow: /0.*/
Disallow: /1.*/
Disallow: /2.*/
Disallow: /3.*/
Disallow: /4.*/
Disallow: /5.*/
```

LindsayHill:
[...]

PR #443, opened 2017-03-30
add support for canonical URL

LindsayHill:
Added support for adding links like this into each page `<link rel="canonical" href="http://docs.stackstorm.com/reference/proxy.html"/>`

This will help steer Google towards using the current version of the docs, rather than a specific version. This addresses #403 in part.

Also changed our `base_url` to https://docs.stackstorm.com/, because we are redirecting users to https.

Looks good in local testing, if it goes OK with /latest, I'll back port to 2.x branches (and I still need to figure out how to rebuild 1.x doc branches to update them too).

### DIFF
--- a/docs/source/_themes/sphinx_rtd_theme/layout.html
+++ b/docs/source/_themes/sphinx_rtd_theme/layout.html
@@ -26,6 +26,10 @@
   {% endblock %}
 
   <link rel="shortcut icon" href="{{ pathto('_static/favicon.ico', 1) }}?v4"/>
+  {# CANONICAL URL #}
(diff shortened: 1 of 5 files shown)


LindsayHill:
Build is failing when building bwc docs, because it does not yet have https://github.com/StackStorm/ipfabric-docs/pull/78 merged. I _think_ it will build OK once we merge that. If someone gives a +1 on the other PR, we can merge that, then re-run these tests, and it _should_ pass.

LindsayHill:
Discussed with Alex, looks like this will work for us. Have now merged ipfabrics PR, so tests are passing again here.

PR #446, opened 2017-03-31
missed http -> https

LindsayHill:
Missed one of the http - > https transitions. This affects the Canonical URL change


Kami:
Good catch :+1:

PR #450, opened 2017-03-31
2.0 Canonical URLs

LindsayHill:
Steer Google towards latest docs. Partly addresses #403


LindsayHill:
Wrong base. Ignore this.

PR #447, opened 2017-03-31
added runner_type, api_key, key_value_pair API auto-gen

LindsayHill:
Updated auto-gen script to include generation of API keys, Key-value pairs and runner_types permissions.

Needs https://github.com/StackStorm/st2/issues/3324 merged before tests will pass.

Addresses https://github.com/StackStorm/ipfabric-docs/issues/77

PR #452, opened 2017-04-01
added runner_type, api_key, key_value_pair API auto-gen

LindsayHill:
Same as #447, but for v2.2


LindsayHill:
Needs this one merged https://github.com/StackStorm/st2/pull/3325 before tests will pass

PR #455, opened 2017-04-04
Add universal version picker and version warning

enykeev:
Universal version picker supposed to show the same set of versions no matter which version of documentation you're currently viewing. Previously, the version picker only showed current version to be the latest one.

Also, there's now a warning on every outdated page pointing to the latest version of documentation.

### DIFF
--- a/docs/source/_themes/sphinx_rtd_theme/layout.html
+++ b/docs/source/_themes/sphinx_rtd_theme/layout.html
@@ -197,7 +197,7 @@
   </script>
   {% endif %}
 
-  <script type="text/javascript" src="/{{ pathto('_static/js/versions.js', 1) }}"></script>
+  <script type="text/javascript" src="{{ pathto('/_static/js/versions.js', 1) }}"></script>
 
   {%- block footer %} {% endblock %}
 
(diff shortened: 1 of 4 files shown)


Kami:
> Universal version picker supposed to show the same set of versions no matter which version of documentation you're currently viewing. 

Agreed.

Kami:
Didn't build and test it locally yet, but LGTM :+1:

Kami:
What else (if anything) needs to be done so we can merge this?

enykeev:
I feel like I'm missing something crucial here but I don't see what exactly. I'm afraid we would not be able to spot it until we merge and rebuild docs for every version we care about.

LindsayHill:
I was hoping to test it locally, but I haven't had a chance. Not even sure if it would work testing it locally. Might be worth just merging it to master now, and check to see how it shows up on `/latest/`

enykeev:
Problem with the change that it won't show up on latest until you build it for both latest and stable. Stable one (`/`) will always be the one loading on every page. Other than that, yeah, go ahead, I don't have a better plan.

LindsayHill:
When I've got a couple of dedicated hours, I'll YOLO-merge it against master & 2.2, and then if that looks OK, push it out to other branches. Will be no later than end of next week. Will do it this week if I can, but connectivity is a bit intermittent.

Issue #457, opened 2017-04-04
Need to restart chatops to delete duplicate aliases


Mierdin:
A user [was noticing duplicate chatops aliases](https://stackstorm-community.slack.com/archives/C066APT88/p1491333849029639) when they made changes to them - apparently `st2chatops` caches each iteration of these. 

Turns out, restarting `st2chatops` reloads these from the filesystem, so we should add a note to the docs that highlights this.

Maybe consider adding here: https://docs.stackstorm.com/chatops/aliases.html#removing

estee-tew:
https://stackstorm.slack.com/archives/community/p1491334304217764

PR #458, opened 2017-04-05
Removing aliases needs st2chatops restart

LindsayHill:
Closes https://github.com/StackStorm/st2docs/issues/457

### DIFF
--- a/docs/source/chatops/aliases.rst
+++ b/docs/source/chatops/aliases.rst
@@ -71,7 +71,8 @@ To remove an action alias, use:
 
    st2 action-alias delete {ALIAS}
 
-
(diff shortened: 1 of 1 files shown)


docs/source/chatops/aliases.rst
```diff
@@ -71,7 +71,8 @@ To remove an action alias, use:
 
    st2 action-alias delete {ALIAS}
 
-
+After removing aliases, you may need to restart the ``st2chatops`` service, or the old commands may still show up
```
Mierdin: Might want to also mention this should happen with changes, since the original reporter of this was making changes to the YAML definitions and that's where the duplicates were coming from.

PR #459, opened 2017-04-07
Documentation for secrets masking in API response and log messages functionality (and related limitations)

Kami:
We initially mostly implemented this feature for one of our customers and we were aware it had (many) limitations so we didn't document it.

We of course agreed on those limitations (cc @dzimine @lakshmi-kannan). Some of them could be solved in a realistic time frame and amount of code (e.g. schema for action results and ability to mark specific attribute in the result as a secret), but others would simple require a lot of work for a very little value and quite a large performance overhead. This would basically be a huge maintenance burden and would significantly increase amount of code which of course means increased surface area for bugs and potential security related issues (less is more and better, always).

@enykeev encountered some issues while working on the new API v2 because we didn't have enough extensive tests for it (good catch, thanks :+1:). Now it's the time we actually document this feature and its limitations (we will of course also make sure we have good extensive test coverage in st2 repo which would allow us to catch such issues earlier and automatically).

As I always say, feature either exists and it's well documented and tested or it doesn't. Right now we are in the realm of "exists with no docs and little tests" so the goal of this PR is to fix that.


docs/source/reference/secrets_masking.rst
```diff
@@ -0,0 +1,146 @@
+Secrets Masking
+---------------
+
+|st2| offers functionality for masking secrets in API responses and log messages which is enabled
+by default.
+
+If you want to disable it, you can do so by setting setting manipulating ``api.mask_secrets`` and
```
LindsayHill: Pick just one out of 'setting setting manipulating'

Kami: Thanks, will fix it :D

PR #465, opened 2017-04-13
Fix typos in action runners

NikonNLG:


Kami:
Thanks, good catch.

The changes look good to me.

PR #469, opened 2017-04-26
Update "coordination" of st2.conf in Policy.rst

tanishige:
Need ':' before password in redis url schema.
https://redis-py.readthedocs.io/en/latest/#redis.StrictRedis.from_url
For example:
redis://[:password]@localhost:6379/0


Kami:
Merged.

Good catch - thanks!

PR #485, opened 2017-05-05
Added robots.txt

LindsayHill:
We've made several changes with canonical URLs and redirect files, to steer Google away from indexing old content. This is working pretty well - almost all searches are now returning the 'right' results.

This change adds robots.txt, so Google won't show version-specific entries below the 'right' answers.

### DIFF
--- a/docs/source/_redirects/robots.txt
+++ b/docs/source/_redirects/robots.txt
@@ -1,0 +1,7 @@
+User-agent: *
+Disallow: /0.*/
+Disallow: /1.*/
+Disallow: /2.*/
+Disallow: /3.*/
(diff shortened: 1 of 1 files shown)


Kami:
:+1:

LindsayHill:
@armab the canonical URLs has made a real difference. I've also pushed some other changes for redirects that should resolve some of the last issues related to things like /config/index vs /install/config. 

Once this change has been indexed by Google, we should see those other entries disappear. Success at last!

Only remaining item is the version warning, which is pushed to v2.1 and /latest now. Looks OK for 2.1, but doesn't work quite right for 2.2 yet. More digging needed :)

PR #494, opened 2017-05-09
CircleCI initial yml

LindsayHill:
Initial `circle.yml`, for migrating from Travis to CircleCI, to standardise with other ST2 projects.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -23,5 +23,5 @@ machine:
 
 test:
   override:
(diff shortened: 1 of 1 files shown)


LindsayHill:
OK, that didn't like having just `make requirements` in the dependencies section, since it also relies upon `clone-st2` running. Have removed dependencies step for now, as it looks like Makefile needs a bigger overhaul

PR #512, opened 2017-06-02
Parallel make steps

LindsayHill:
Split `make docs` and `make bwcdocs` to improve test time

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -23,5 +23,5 @@ machine:
 
 test:
   override:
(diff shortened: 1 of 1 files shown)


LindsayHill:
Results look good - builds get run on two containers - see https://circleci.com/gh/StackStorm/st2docs/57

Previous run times were 5-6 mins. This change took 3:32. Small improvement, but hey, it's a win.

PR #517, opened 2017-06-09
Datastore values must be string. Fixes #510

### LINKED ISSUE #510: Clarify datastore requirements
It seems that the datastore requires that all values be strings. Since it is backed by mongo, I figured it would be happy to store json objects as well. It feels funny to do a json.dumps to store a json object as a string in what is ultimately backed by a json-native document store.

Maybe clarify in the docs about what data types are allowed for datastore values, and hopefully explain why those requirements are there.

LindsayHill:
Note that datastore values must be a string.

Fixes #510

### DIFF
--- a/docs/source/datastore.rst
+++ b/docs/source/datastore.rst
@@ -16,6 +16,16 @@ rule evaluation.
 Key-Value pairs can also have a TTL associated with them, for automatic
 expiry. 
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for clarifying this.

IIRC, there is also `get_object` / `set_object` or a similarly named action in `st2` pack which transparently handles serialization and de-serialization for a user.

PR #519, opened 2017-06-09
note on min requirement for /var and /opt.

### LINKED ISSUE #501: System requirements should clarify partition sizing requirements
Current system requirements list total disk size for test & production systems. This works for users that have one large partition, but can cause problems for those with complicated partitioning schemes. We should include some minimum sizes for partitions such as `/opt/` and `/var`. 

Note that rabbitmq & mongodb require certain minimum amounts of free space, or they will not operate correctly.

LindsayHill:
Fixes #501


Kami:
LGTM :+1:

PR #522, opened 2017-06-13
Added step to upgrade docs about upgrading the installed packs

nmaludy:
The upgrade docs lacked a step about upgrading the installed packs.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -6,6 +6,13 @@ Upgrade Notes
 |st2| in development
 --------------------
 
+* The API endpoint for searching or showing packs has been updated to return an empty list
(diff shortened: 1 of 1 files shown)


LindsayHill:
I'm trying to decide if we should have this in this doc, which lists generic upgrade steps, or if we should just have a version-specific entry under https://docs.stackstorm.com/upgrade_notes.html?

I'm also not sure if we should recommend upgrading all packs, or just some?

Maybe we should include something in the generic upgrade steps to say "this is a great time to check your packs are up to date", and then in the version-specific upgrade_notes, call out any packs that *must* be upgraded.

FWIW, this is going to become an issue in future for more packs too - we will soon merge code that enforces uniqueness across all positional params. I've fixed a couple of packs in Exchange, but not everyone will be using the latest packs.

PR #527, opened 2017-06-17
Fixed upgrade notes

LindsayHill:
Additional v2.2 heading in there, same as https://github.com/StackStorm/st2docs/pull/526, for v2.3 branch


Kami:
Good catch :+1:

PR #542, opened 2017-07-05
New Jinja Filters, Structure Revamp of Whole Page

Mierdin:
- Update documentation for new filters introduced in https://github.com/StackStorm/st2/pull/3480 and https://github.com/StackStorm/st2/pull/3482
- Restructure the page to flow a bit better


Mierdin:
@armab You're right, though I'm going to look into getting them added to st2mistral today - maybe long-term the right approach is to add filters to both, removing this restriction?

Mierdin:
@armab The st2mistral work is going to take a bit more time, so in the meantime I just added note and will create an issue in st2mistral to update this when that's done

docs/source/reference/jinja.rst
```diff
@@ -1,101 +1,87 @@
...
+to allow some advanced capabilities in working with variables.
+
+.. _referencing-datastore-keys-in-jinja:
+
+Custom Jinja Filters
+--------------------
+
+In addition to `standard filters <http://jinja.pocoo.org/docs/dev/
+templates/#builtin-filters>`_ available in Jinja, |st2| supports custom filters
+as well.
```
LindsayHill: "In addition to...supports custom filters as well" - feels like we don't need "as well" at the end, as it's redundant?

docs/source/reference/jinja.rst
```diff
@@ -1,101 +1,87 @@
...
+For brevity, only simple Jinja patterns for each filter are documented below. "Real-world" usage
+will depend on the type of content where the filters are being applied (sensors, triggers, rules,
+action and workflows) and their syntax. More detailed examples can be found in several of the
+ActionChains located in the ``examples`` pack:
+:github_st2:`st2/contrib/examples/actions/chains/ </contrib/examples/actions/chains/>`.
 
-.. _jinja-jinja-filters:
+..  TODO We should consider separating each specific usage into individual ActionChains and refer to
+    it using literalinclude (i.e. .. literalinclude:: /../../st2/contrib/examples/actions/workflows/mistral-jinja-branching.yaml)
+    so we can just use the code as the source of truth. Then, we can remove the above note.
```
LindsayHill: Seems reasonable to do it with literalinclude

Mierdin: Cool. To keep things easy, we'd have to split up the existing chains into one example per workflow (not to mention some of the filters just outright don't have examples) so I'm thinking we just leave the TODO in here for now until the examples pack is updated. Opened https://github.com/StackStorm/st2/issues/3553 to track this

docs/source/reference/jinja.rst
Mierdin: Sounds good, thanks

docs/source/reference/jinja.rst
```diff
@@ -1,101 +1,86 @@
...
+~~~~~~~~~~
+Decrypt a system scoped datastore item.
+
+.. code-block:: bash
+
+    {{value_key | decrypt_kv }}
+
+json_escape
+~~~~~~~~~~~
+Adds escape characters to JSON strings.
```
Kami: I think it would also be good to add examples of inputs and outputs for some filters to make it more obvious what they do.

docs/source/reference/jinja.rst
```diff
@@ -1,101 +1,86 @@
...
-.. _jinja-jinja-filters:
+..  TODO We should consider separating each specific usage into individual ActionChains and refer to
+    it using literalinclude (i.e. .. literalinclude:: /../../st2/contrib/examples/actions/workflows/mistral-jinja-branching.yaml)
+    so we can just use the code as the source of truth. Then, we can remove the above note.
 
-Custom Jinja Filters
---------------------
+.. warning::
+
+    These custom filters are currently not available in Mistral workflows. To use them in a workflow, you must use ActionChains.
```
armab: :+1:

PR #543, opened 2017-07-05
Add path to st2sensorcontainer

LindsayHill:


Kami:
Good catch :+1: 

Will cherry pick into v2.3

PR #544, opened 2017-07-05
Fixed typo in RBAC mappings directory

LindsayHill:


Kami:
Good catch :+1: 

Will cherry pick into v2.3

PR #547, opened 2017-07-10
Fixing reference to regex_replace filter in jinja.

Xorso:
Need to escape two \'s in order for the filter to work.


Kami:
Good catch, thanks :+1:

LindsayHill:
Looks like you also need to escape the backslash when used in a regular text field in RST, so added another level of escaping.

PR #552, opened 2017-07-15
Updates to custom YAQL and Jinja filters

Mierdin:
Covers changes in https://github.com/StackStorm/st2mistral/pull/30

- Generally covers that the filters are now available in Mistral)
- Breaking change to `st2kv` that adds a parameter to explicitly enable decryption in st2kv (disabled by default now).
- Adds note about breaking change to upgrade notes


docs/source/upgrade_notes.rst
```diff
@@ -17,6 +17,10 @@ Upgrade Notes
   instead of ``None`` when the pack was not found in the index. This is technically a breaking
   change, but a necessary one because returning ``None`` caused the client to throw an exception.
 
+* The ``st2kv`` function has been changed (in 2.3.2) so that it no longer attempts to decrypt stored values by
+  default. To return decrypted values, this must be explictly enabled via parameter, like so:
+  ``st2kv('st2_key_id', decrypt=true)``.
```
armab: This looks like a breaking change.

I think we should include it in major release like `2.4.0`, instead of `2.3.2` patch release.
Let's be consistent, from the patch release users expect bug fixes, not changing the important core behavior and existing workflows they rely on.

cc @m4dcoder @lakshmi-kannan

Mierdin: I personally don't mind either way. Only reason I was aiming for 2.3.2 is because @lakshmi-kannan recommended it. I'll let him chime in on the matter; there's no question that this is a breaking change.

Mierdin:
@armab @lakshmi-kannan FYI I removed references to 2.3.2, updated to 2.4

PR #554, opened 2017-07-18
Small typo

lingfish:
Proxies need to use http URIs, or this happens:

`TASK [st2repo : Add keys to keyring] ******************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Unsupported proxy scheme: https. Currently ansible only supports HTTP proxies."}`


armab:
Thanks!

Seems there is still an issue opened: https://github.com/ansible/ansible/issues/10941

PR #570, opened 2017-08-03
Fixed minor typos

crepetl:


LindsayHill:
Thanks - no matter how many I find, there's always more!

PR #580, opened 2017-08-17
Change log configuration path

LindsayHill:
Change to newer style log configuration path in PAM example

### DIFF
--- a/docs/source/authentication.rst
+++ b/docs/source/authentication.rst
@@ -76,7 +76,7 @@ repo. The following is a sample auth section in the config file for the PAM back
     use_ssl = True
     cert = /path/to/ssl/cert/file
     key = /path/to/ssl/key/file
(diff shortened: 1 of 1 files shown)


docs/source/authentication.rst
```diff
@@ -76,7 +76,7 @@ repo. The following is a sample auth section in the config file for the PAM back
     use_ssl = True
     cert = /path/to/ssl/cert/file
```
nmaludy: Should we maybe put the default path to the cert and SSL key file in here also?

LindsayHill: Thought about it; decided not to. This SSL cert could be different to the default self-signed cert.

nmaludy: :+1:

LindsayHill:
Related: https://github.com/StackStorm/st2-auth-backend-pam/pull/12

PR #582, opened 2017-08-18
Initial draft of the "Docker Image" documentation

warrenvw:
It would be great if we can get one or two folks who haven't previously used Docker (or at least `st2-docker`) to read through the documentation and see if the instructions make sense and examples in the Sample Usage work as expected.

Remaining tasks:
 - [ ] Read through the document and ensure everything that needs to be said has been said.
 - [ ] Ensure all the steps work as expected.


armab:
Looks good, but IMO this all could be fine-polished into a smaller piece of text with more code specifics and facts. I think about it as a technical doc and not as blog post.

For example, big `Requirements` paragraph could be shortened to just minimal docker engine version, minimal docker-compose version and docker-compose file format version.

But it's just personal preference and not a request to rewrite it, so take with a grain of salt.

dzimine:
Warren, let's reduce the amount of info here to the minimal, 
and refer the reader to https://github.com/StackStorm/st2-docker/blob/master/README.md for the up-to-date information on our actively growing docker.

LindsayHill:
I think that the installation & requirements info belongs here, but the other stuff (e.g. running actions, adding rules, etc) doesn't. We don't put that info on the install page for VM environments, we have it in other sections. Unless there's something specifically different about usage here, we don't need to call it out.

I'm fine with referring to a README on the Docker page for more specific Docker usage info, **but**: we can't just have a minimal couple of lines here with a pointer to a README - if we do, that will make this look like a second-class citizen.

We may also want to update some of the other pages, e.g. docs.stackstorm/install/index, and docs.stackstorm.com/install/system_requirements.html

docs/source/install/index.rst
```diff
@@ -73,6 +73,7 @@ For more detail on reference deployments, or OS-specific manual installation ins
 
     Reference Deployment Overview <overview>
     system_requirements
+    Docker Image <docker>
```
armab: In a list of installers, I think it's better to move it after instructions to install on U16/EL6/EL7 since they have more priority/adoption.

Also `Docker` might look better, comparing to `Docker Image(s)`.

armab: Also `Docker` might look better, comparing to `Docker Image`, because we might think about it as Docker Image (all-in-one), but there are also Docker Images (1 service per container) and someone might think about Docker Container(s).
So just `Docker` covers all that.

LindsayHill:
Good call from @armab on using "Docker" vs "Docker Image" - looks a bit cleaner.

LindsayHill:
@warrenvw you now have two approvals, and checks are passing. Time to merge this?

PR #587, opened 2017-08-22
Update the Mistral DB upgrade to exclude errors from OpenStack actions

m4dcoder:
On Mistral DB upgrade, the populate commands throw errors when registering the native OpenStack actions. Modify the command to exclude the OpenStack related errors.


Mierdin:
@m4dcoder 👍 Thank you.

(related to https://github.com/StackStorm/st2-packages/pull/486 which covers installs but not upgrades)

PR #620, opened 2017-09-30
Overhauled install section

LindsayHill:
Re-worked the installation section, in an attempt to make it clearer to users what their installation options are. 

Lots of changes here, and probably a bit hard to review in this format. Suggest that it's probably better to merge them, then check docs.stackstorm.com/latest/install, and see how that looks.


docs/source/install/common/datastore_crypto_key.rst
```diff
@@ -1,23 +1,22 @@
...
+  sudo st2-generate-symmetric-crypto-key --key-path ${DATASTORE_ENCRYPTION_KEY_PATH}
 
-    # Make sure only st2 user can read the file
-    sudo usermod -a -G st2 st2
-    sudo chgrp st2 ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
-    sudo chmod o-r ${DATASTORE_ENCRYPTION_KEYS_DIRECTORY}
-    sudo chgrp st2 ${DATASTORE_ENCRYPTION_KEY_PATH}
-    sudo chmod o-r ${DATASTORE_ENCRYPTION_KEY_PATH}
+  # Make sure only st2 user can read the file
+  sudo usermod -a -G st2 st2
```
armab: For some unknown reason it's included here.

`st2` user is already part of `st2` group, added by [`deb`](https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/preinst#L36) and [`rpm`](https://github.com/StackStorm/st2-packages/blob/master/packages/st2/rpm/preinst_script.spec#L21) packaging.

We can remove this line.

LindsayHill: Removed that line

PR #624, opened 2017-10-05
Update pip & Sphinx versions, allow page-specific Github URL

LindsayHill:
Match pip version used for st2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ requirements: virtualenv
 	@echo
 
 	# Make sure we use latest version of pip
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=8.1.2,<8.2"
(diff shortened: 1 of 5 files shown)


LindsayHill:
Might also look at updating Sphinx version as part of this PR. Hold off merging.

LindsayHill:
Want to do some more testing with updated Sphinx versions before merging.

LindsayHill:
Added a couple more things here

 - Fixed up the double-slash that appears in the GitHub URL, by separating out GitHub user + repo
- Added option to over-ride GitHub URL through page-specific metatada
- Updated Sphinx version

LindsayHill:
Looks like https://github.com/StackStorm/ipfabric-docs/pull/112 needs to be merged before these tests will pass. Specifically, it needs github_user defined in info.py for BWC docs

PR #628, opened 2017-10-09
New rule operators inside and ninside

lampwins:
Corresponding to https://github.com/StackStorm/st2/pull/3781

### DIFF
--- a/docs/source/rules.rst
+++ b/docs/source/rules.rst
@@ -199,6 +199,9 @@ This section describes all the available operators which can be used in the crit
                   greater than the provided value.
 ``exists``        Key exists in payload.
 ``nexists``       Key doesn't exist in payload.
+``inside``        Trigger value is inside provided value. Reverse of `contains`.
+``ninside``       Trigger value is not inside provided value. Reverse of
+                  `ncontains`.
 ================= =================================================================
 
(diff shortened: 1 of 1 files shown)


docs/source/rules.rst
```diff
@@ -199,6 +199,9 @@ This section describes all the available operators which can be used in the crit
                   greater than the provided value.
 ``exists``        Key exists in payload.
 ``nexists``       Key doesn't exist in payload.
+``inside``        Trigger value is inside provided value. Reverse of `contains`.
```
Kami: I would add some more explanation to make it more clear and obvious since it might still be a bit confusing what's the difference between those operations.

Maybe something along the lines of "Reverse of contains (contains checks for provided value aka pattern inside the trigger attribute and inside checks for trigger value inside the provider value aka pattern)."

Kami: Also, I know it's a bit wordy, but as I said above, I think it's a bit confusing (and easy to mix those operators up) without additional clarification and examples :)

LindsayHill:
Will merge this once https://github.com/StackStorm/st2/pull/3781 gets merged

Kami:
Thanks.

PR #630, opened 2017-10-12
Adjust docs for 'resultstracker.query_interval' and Mistral completion latency & CPU overhead

armab:
Changing the `resultstracker.query_interval` doc setting from `20s` to `5s`, following the https://github.com/StackStorm/st2/pull/3776.

So seems like @lakshmi-kannan already included good explanation in `docs` about the Mistral latency troubleshooting when implemented `resultstracker.query_interval` setting.

### DIFF
--- a/docs/source/mistral.rst
+++ b/docs/source/mistral.rst
@@ -296,3 +296,13 @@ There are more workflow examples under :github_st2:`/usr/share/doc/st2/examples 
 Check out this step-by-step tutorial on building a workflow in |st2| https://stackstorm.com/2015/07/08/automating-with-mistral-workflow/
 
 More details about Mistral can be found at https://docs.openstack.org/mistral/latest/.
+
+
+Mistral Workflows Latency and Performance
+-----------------------------------------
+|st2| interacts with Mistral via HTTP APIs. This applies when kicking off a workflow execution
+or collecting the results for a running workflow. |st2| queries Mistral to check on workflow
+execution status and the status of individual tasks. This mechanism has a number of configuration settings.
+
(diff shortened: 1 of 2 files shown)


docs/source/troubleshooting/mistral.rst
```diff
@@ -21,6 +21,7 @@ When that happens, the new database tables, relationships, and indices must be d
 ``mistral-db-manage upgrade head`` command can be re-run. For more details, review the specific
 section for the version being upgraded in :doc:`/install/upgrades`.
 
+.. _mistral-workflows-latency:
```
armab: Linking from the mistral page to mistral troubleshooting page.

@LindsayHill Please take a look once again if this construct something that'll work. Took an example from another page.

docs/source/mistral.rst
```diff
@@ -296,3 +296,13 @@ There are more workflow examples under :github_st2:`/usr/share/doc/st2/examples
...
 More details about Mistral can be found at https://docs.openstack.org/mistral/latest/.
+
+
+Mistral Workflows Latency and Performance
+-----------------------------------------
+|st2| interacts with Mistral via HTTP APIs. This applies when kicking off a workflow execution
+or collecting the results for a running workflow. |st2| queries Mistral to check on workflow
+execution status and the status of individual tasks. This mechanism has a number of configuration settings.
+
+See :ref:`mistral-workflows-latency` section about how to fine-tune
```
armab: And here

docs/source/troubleshooting/mistral.rst
LindsayHill: Yes, that's the right sort of construct to use for cross-reference

PR #633, opened 2017-10-16
Add a note about the installation behind a proxy

armab:
Following the https://github.com/StackStorm/st2-packages/pull/510 implementation, adding a note about installing StackStorm with `curl|bash` installer behind a proxy.

Example: 
```sh
export http_proxy=http://192.168.10.123:3128
export https_proxy=http://192.168.10.123:3128
export no_proxy=127.0.0.1,localhost

curl -sSL https://stackstorm.com/packages/install.sh |  bash -s -- --user=st2admin --password=st2admin
```


armab:
Included export proxy example in https://github.com/StackStorm/st2docs/pull/633/commits/d6db6c646ebd76600af29c530d57867148e8c026

PR #641, opened 2017-10-24
Update index.rst

mdsarmie:
Updated the introduction statement

### DIFF
--- a/docs/source/troubleshooting/index.rst
+++ b/docs/source/troubleshooting/index.rst
@@ -1,8 +1,8 @@
 Troubleshooting
 ===============
 
-This section includes common problems you might encounter while using the platform,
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting/index.rst
```diff
@@ -1,8 +1,8 @@
 Troubleshooting
 ===============
 
-This section includes common problems you might encounter while using the platform,
-and how to troubleshoot and debug them.
+This section includes common problems you might encounter while using the Brocade Workflow Composer platform,
```
LindsayHill: I think we should use `|fullname|` here, rather than "Brocade Workflow Composer", so it will render properly for StackStorm vs BWC docs.

PR #642, opened 2017-10-24
Update self_verification.rst

mdsarmie:
Minor updates made for better descriptions.

### DIFF
--- a/docs/source/troubleshooting/self_verification.rst
+++ b/docs/source/troubleshooting/self_verification.rst
@@ -1,7 +1,7 @@
 Running Self-Verification
 =========================
 
-|st2| includes a script for verifying the BWC installation.
+|st2| includes a script for verifying the system installation.
 
 The script covers the following aspects of |st2|:
 
@@ -14,7 +14,7 @@ The script covers the following aspects of |st2|:
 To run the |st2| self-verification script:
 
 1. If you don't have :ref:`encryption keys setup already<admin-setup-for-encrypted-datastore>`, do so.
-   This will require both |st2| and system admin privileges when setting it up.
+   This will require both |st2| and system admin privileges to configure.
 
 2. Run the self-check script ``/opt/stackstorm/st2/bin/st2-self-check``. This also copies the examples from ``/usr/share/doc/st2/examples`` to ``/opt/stackstorm/packs/`` and registers the content. Running this step will pollute your |st2| environment because it will download fixtures from `st2tests <https://github.com/StackStorm/st2tests/tree/master/packs/>`__.
 
--- a/docs/source/troubleshooting/self_verification.rst
+++ b/docs/source/troubleshooting/self_verification.rst
@@ -1,7 +1,7 @@
 Running Self-Verification
 =========================
 
-|st2| includes a script to verify the installation, using |st2| itself.
+Brocade Workflow Composer includes a script for verifying the BWC installation.
 
 The script covers the following aspects of |st2|:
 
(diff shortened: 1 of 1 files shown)


docs/source/troubleshooting/self_verification.rst
```diff
@@ -1,7 +1,7 @@
 Running Self-Verification
 =========================
 
-|st2| includes a script to verify the installation, using |st2| itself.
+Brocade Workflow Composer includes a script for verifying the BWC installation.
```
LindsayHill: The docs build process gets a bit complicated, but there's a few things going on here.

One is that we use the same base documentation for docs.stackstorm.com and bwc-docs.brocade.com. The difference is that the bwc-docs have a 'Brocade' theme applied, and the Automation Suites documentation gets merged in.

Another key difference is that `|st2|` gets automatically replaced by "StackStorm" on docs.stackstorm.com, and by "BWC" on bwc-docs.brocade.com. `|fullname|` gets replaced by "StackStorm" on docs.stackstorm.com, and "Brocade Workflow Composer" on bwc-docs.

In general, we want to refer to the product as "StackStorm" on docs.stackstorm.com, and "BWC/Brocade Workflow Composer" on bwc-docs.brocade.com. So in this specific case, you should use either `|st2|` or `|fullname|`.

docs/source/troubleshooting/self_verification.rst
```diff
@@ -11,14 +11,14 @@ The script covers the following aspects of |st2|:
...
+To run the BWC self-verification script:
 
 1. If you don't have :ref:`encryption keys setup already<admin-setup-for-encrypted-datastore>`, do so.
-   This requires both |st2| and system admin privileges.
+   This will require both |st2| and system admin privileges when setting it up.
 
-2. Run the self-check script. This also copies the examples from ``/usr/share/doc/st2/examples``
-   to ``/opt/stackstorm/packs/`` and registers the content. This step pollutes your |st2| environment by
-   downloading fixtures from `st2tests <https://github.com/StackStorm/st2tests/tree/master/packs/>`__.
+2. Run the self-check script "/opt/stackstorm/st2/bin/st2-self-check". This also copies the examples from ``/usr/share/doc/st2/examples``
```
LindsayHill: I would prefer it if we used double backticks around command names, so it renders as `fixed-width`.

Looks like we also need to fix up the indentation for the `.. code-block:: bash` section below. I think we should indent it, so it shows up as part of list item 2 here, rather than a new paragraph

docs/source/troubleshooting/self_verification.rst
```diff
@@ -11,14 +11,14 @@ The script covers the following aspects of |st2|:
...
+   to ``/opt/stackstorm/packs/`` and registers the content. Running this step will pollute your |st2| environment because it will download
+   ixtures from `st2tests <https://github.com/StackStorm/st2tests/tree/master/packs/>`__.
```
LindsayHill: s/ixtures/fixtures/

docs/source/troubleshooting/self_verification.rst
mdsarmie: Thank you. Update made per comment. See new commit changes.

docs/source/troubleshooting/self_verification.rst
mdsarmie: Fixed.

docs/source/troubleshooting/self_verification.rst
mdsarmie: Fixed

PR #665, opened 2017-11-09
Add a few helpful hints for troubleshooting webhooks

Mierdin:


armab:
Nicely composed doc :+1: 

However I think the good place for it is Troubleshooting section (https://docs.stackstorm.com/troubleshooting/index.html) and link to that page from the `webhooks`.

docs/source/webhooks.rst
```diff
@@ -172,6 +172,43 @@ To list all registered webhooks, run:
 
     st2 webhook list
 
+My Webhook Isn't Working!
+-------------------------
+
+Often you may run into issue with getting StackStorm to respond to webhooks the way you want. A common problem
```
LindsayHill: I would prefer it if we did not use "Often you may..." here. Just say something like "You may run into issues with getting |st2| to respond to webhooks the way you want..."

docs/source/webhooks.rst
```diff
@@ -172,6 +172,43 @@ To list all registered webhooks, run:
...
+that community members run into is that StackStorm doesn't seem to recognize incoming webhooks, even if a manual
+test using ``curl`` results in a successful HTTP status.
+
+The first thing to do is confirm that the webhook request is indeed getting to StackStorm. When webhooks successfully
```
LindsayHill: Where possible, use |st2| rather than "StackStorm" - we have some macros for this.

docs/source/webhooks.rst
```diff
@@ -172,6 +172,43 @@ To list all registered webhooks, run:
...
+    2017-11-09 21:21:00,563 140040790988752 INFO logging [-] 651897d7-2aed-4a11-9c14-4c56152230cc - POST /v1/webhooks/st2 with query={} (remote_addr='127.0.0.1',method='POST',request_id='651897d7-2aed-4a11-9c14-4c56152230cc',query={},path='/v1/webhooks/st2')
+    2017-11-09 21:21:00,566 140040790988752 AUDIT auth [-] Token with id "5a04be6bc4da5f0d1fe22ca1" is validated.
+    2017-11-09 21:21:00,572 140040790988752 INFO logging [-] 651897d7-2aed-4a11-9c14-4c56152230cc - 202 94 8.623ms (content_length=94,request_id='651897d7-2aed-4a11-9c14-4c56152230cc',runtime=8.623,remote_addr='127.0.0.1',status=202,method='POST',path='/v1/webhooks/st2')
+
+If you don't see these, the actual request isn't even making it to StackStorm. You should look at anything along the path between the requester and StackStorm and ensure that nothing is blocking that communication. Don't forget to look at the NGINX logs as well, as this front-ends all API communication, including webhooks, in a normal setup.
+
+You may also be running StackStorm with a self-signed certificate. Be sure that the webhook sender isn't rejecting the connection because of this, as that will prevent the webhook from ever being sent.
```
LindsayHill: I wonder if we should offer guidance on how to detect this? Guess we say "you won't see connections in nginx logs"?

Mierdin:
@LindsayHill @armab Thanks for your comments. I should have addressed them in the last few commits.

PR #669, opened 2017-11-22
Clarified upgrade instructions

LindsayHill:
Trying to make it simpler for users to upgrade.

### DIFF
--- a/docs/source/install/upgrades.rst
+++ b/docs/source/install/upgrades.rst
@@ -28,13 +28,13 @@ This is the standard upgrade procedure:
 
    .. sourcecode:: bash
 
-      sudo apt-get install --only-upgrade st2 st2web st2chatops st2mistral bwc-enterprise
+      sudo apt-get install --only-upgrade st2 st2web st2chatops st2mistral
 
    RHEL/CentOS:
 
    .. sourcecode:: bash
 
-      sudo yum update st2 st2web st2chatops st2mistral bwc-enterprise
+      sudo yum update st2 st2web st2chatops st2mistral
 
 3. Upgrade Mistral database:
 
--- a/docs/source/install/upgrades.rst
+++ b/docs/source/install/upgrades.rst
@@ -4,8 +4,9 @@ Upgrades
 When new versions of |st2| are released, they are published to our APT and Yum repositories. You
 can use standard Linux package management tools to install these upgraded packages.
 
-Depending on the versions you are upgrading to and from, you may need to run additional data model
-:ref:`migration scripts<migration-scripts-to-run>`.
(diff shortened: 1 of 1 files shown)


docs/source/install/upgrades.rst
```diff
@@ -4,37 +4,37 @@ Upgrades
...
-2. Upgrade |st2| packages (``st2``, ``st2web``, ``st2chatops``, ``st2mistral`` and
-   ``bwc-enterprise`` using distro-specific tools:
+2. Upgrade |st2| packages using distro-specific tools:
 
    Ubuntu:
 
    .. sourcecode:: bash
 
-      sudo apt-get install --only-upgrade $PKG_NAME
+      sudo apt-get install --only-upgrade st2 st2web st2chatops st2mistral bwc-enterprise
```
m4dcoder: Maybe add a note to say that non-BWC install don't need to include bwc-enterprise?

LindsayHill: I thought about that, but the `-only-upgrade` flag says “only upgrade if this is already installed.”

This way makes for less info for user to read, and should Just Work

m4dcoder: +1

LindsayHill: Actually I've looked a bit more closely at this one. The `--only-upgrade` flag works pretty well...as long as `apt` knows what about that package. So the above commands will fail on systems that don't have the StackStorm Enterprise repo.

Guess I'll need to come up with a different way of phrasing it. I still think it's right to put all the packages in one line. 

It's possible that we don't need to specify bwc-enterprise - it may get picked up by other dependencies if you specify st2web. I'll need to check.

LindsayHill: Looks like we'll be OK if we don't specify `bwc-enterprise`. For systems that don't have it, it won't matter. For systems that do have an older version installed, it will get picked up by apt dependency handling, e.g.:
```sh
brocade@BWC:~$  sudo apt-get install --only-upgrade st2 st2web st2chatops st2mistral
Reading package lists... Done
Building dependency tree
Reading state information... Done
Skipping st2web, it is not installed and only upgrades are requested.
The following packages were automatically installed and are no longer required:
  libxml2-dev libxslt1-dev libxslt1.1
Use 'apt-get autoremove' to remove them.
The following extra packages will be installed:
  bwc-enterprise bwc-ui st2-auth-ldap st2flow
The following packages will be upgraded:
  bwc-enterprise bwc-ui st2 st2-auth-ldap st2chatops st2flow st2mistral
7 upgraded, 0 newly installed, 0 to remove and 31 not upgraded.
Need to get 61.6 MB of archives.
After this operation, 5,207 kB of additional disk space will be used.
Do you want to continue? [Y/n]
```

PR #690, opened 2018-02-21
Corrected explanation about the functionality to parse parameters which are array of objects type.

userlocalhost:
## Background
The changing of [#3909](https://github.com/StackStorm/st2/pull/3909) in the st2 has not been applied to the st2docs.

## Changing points
This patch corrects followings.

- Added a flag `--auto-dict` to parse the colon separated characters in an array typed parameter as array of objects.
- Added description that `--auto-dict` parameter will be deprecated in near future, that is referred from [the description in the usage of st2client](https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L315-L317).
- Fixed a typo of delimiter of array of objects.

### DIFF
--- a/docs/source/reference/cli.rst
+++ b/docs/source/reference/cli.rst
@@ -412,7 +412,7 @@ like this:
 
 .. sourcecode:: bash
 
-    st2 run mypack.set_interfaces \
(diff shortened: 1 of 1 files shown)


Kami:
Merged, thanks :+1:

PR #691, opened 2018-02-21
Added documentation about the required tooz interface

nmaludy:
Previously the documentation did not specify which features/functionality was required to be implemented by the tooz backend. This made it hard to reason about which backend to choose.

I did some research and code grepping and found that StackStorm only utilizes the `Locking` functionality portion of the tooz interface. https://docs.openstack.org/tooz/latest/user/compatibility.html

Added this to the docs, along with a link to the compatibility matrix so that other can make an informed decision.

### DIFF
--- a/docs/source/reference/policies.rst
+++ b/docs/source/reference/policies.rst
@@ -117,6 +117,11 @@ Other supported coordination backends include:
 * PostgreSQL
 * file (for testing when all the services are running on a single host)
 
+StackStorm utilizes the ``OpenStack Tooz`` library for communicating with the
+coordination backend. The coordination backend must support the ``Locking``
+functionality as defined by the ``Tooz`` interface. Please refrence the
+`OpenStack Tooz compatability page <https://docs.openstack.org/tooz/latest/user/compatibility.html>`_
+for more information what interfaces are implemented by various backends.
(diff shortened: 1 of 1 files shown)


docs/source/reference/policies.rst
```diff
@@ -117,6 +117,11 @@ Other supported coordination backends include:
 * PostgreSQL
 * file (for testing when all the services are running on a single host)
 
+StackStorm utilizes the ``OpenStack Tooz`` library for communicating with the
+coordination backend. The coordination backend must support the ``Locking``
+functionality as defined by the ``Tooz`` interface. Please refrence the
+`OpenStack Tooz compatability page <https://docs.openstack.org/tooz/latest/user/compatibility.html>`_
+for more information what interfaces are implemented by various backends.
```
Kami: Line break here probably wouldn't be a good idea.

Kami:
Thanks :+1:

docs/source/reference/policies.rst
nmaludy: Sorry, a little confused. What do you want changed?

Kami: Just a line break to separate those two paragraphs, imo.

PR #694, opened 2018-02-26
Enhance documentation about the default ansible-st2 login creds

armab:
Fixes StackStorm/ansible-st2#189

- Add a note that default login creds are ``testu:testp``, according to https://github.com/StackStorm/ansible-st2#variables.
- Ask user to change the defaults to a more secure ones.

### DIFF
--- a/docs/source/install/ansible.rst
+++ b/docs/source/install/ansible.rst
@@ -38,6 +38,12 @@ commands:
 
   ansible-playbook stackstorm.yml
 
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


armab:
@candlerb I believe this helps with your documentation improvement request https://github.com/StackStorm/ansible-st2/issues/189

Kami:
Thanks :+1: 

Maybe it wouldn't be a bad idea to use the same set of default credentials we use for st2 installer script?

armab:
It's funny enough that it's using st2 installer script credentials, just from the pre-packaging era. We had another curl|bash script where `testu:testp` was default :)

Need to change the defaults in ansible-st2 only with a major version bump, because it may break logins for someone who used old ones.

PR #698, opened 2018-03-02
Fix workflow language link

apolloliu:
There is no more DSL docs since the commit[1]
and all of them should be updated.

[1]https://github.com/openstack/mistral/commit/57e6936d6caa097bc9a017bc8f5ec7ec341345c1#diff-ee8e0e6e515c327d1f6761aa742bc183


Kami:
Thanks :+1: 

Perhaps we should maybe also link to a specific version and not latest so it doesn't break again in the future when / if it changes.

apolloliu:
@Kami Because we are in master, I think we should keep link with the latest version which may include some significant changes.

Kami:
That's a good point, although keeping it up to date manually in version branches could get a bit expensive / annoying.

apolloliu:
It's not a big deal. Good news is that there are many developers to join and make contributions.

PR #704, opened 2018-03-19
Use -1 for no limit, not 0

LindsayHill:
We changed to use `-n -1` for 'no limit', not `-n 0`.

Attempts to use `-n 0` result in this:
```shell
$ st2 key list -n 0
ERROR: 400 Client Error: Bad Request
MESSAGE: Limit, "0" specified, must be a positive number or -1 for full result set. for url: http://127.0.0.1:9101/v1/keys/?scope=system&decrypt=false&limit=0
```


Kami:
:+1:

Issue #705, opened 2018-03-20
Overhaul Windows Runner documentation


LindsayHill:
Overhaul & update Windows Runner documentation.

Provide better examples of using Windows runners

LindsayHill:
No plans to do this in near future. Have not seen any recent interest in this. Will close for now.

PR #706, opened 2018-03-20
Updated winexe URL, suppress Mistral warnings

LindsayHill:
Better winexe URL for Ubuntu systems, and be more consistent with suppressing spurious Mistral warnings when running `populate` commands.

Related #705


Kami:
:+1:

PR #718, opened 2018-04-10
Remove documentation for --python3 flag (master)

Kami:
Removing documentation added in https://github.com/StackStorm/st2docs/pull/717, due to some issues we found in that feature (see https://github.com/StackStorm/st2/pull/4075).


Kami:
Yeah, even though it's behind a feature flag and disabled by default it's better to hide it until we known it's working fully as it should.

PR #728, opened 2018-04-30
Add register step to general upgrade procedures

LindsayHill:
Some upgrades change APIs, or runners. This causes issues if content is not re-registered. Added a step to the general upgrade procedure to register content.


LindsayHill:
Fixes #653

PR #738, opened 2018-05-08
Add a Vagrant doc

armab:
Closes StackStorm/packer-st2/issues/3

StackStorm Community OVA / Vagrant doc.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -126,14 +126,10 @@ rst_epilog = """
 .. _exchange: https://exchange.stackstorm.org/
 .. |bwc| replace:: Extreme Workflow Composer
 .. |ipf| replace:: IP Fabric Automation Suite
-""" % product_replace
-
-# A string of reStructuredText that will be included at the beginning of every source file that is read
-rst_prolog = """
 .. |br| raw:: html
 
    <br />
-"""
+""" % product_replace
 
 # Show or hide TODOs. See http://sphinx-doc.org/ext/todo.html
 todo_include_todos = True
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -126,9 +126,6 @@ rst_epilog = """
 .. _exchange: https://exchange.stackstorm.org/
 .. |bwc| replace:: Extreme Workflow Composer
 .. |ipf| replace:: IP Fabric Automation Suite
-.. |br| raw:: html
-
-   <br />
 """ % product_replace
 
 # Show or hide TODOs. See http://sphinx-doc.org/ext/todo.html
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -128,6 +128,13 @@ rst_epilog = """
 .. |ipf| replace:: IP Fabric Automation Suite
 """ % product_replace
 
+# A string of reStructuredText that will be included at the beginning of every source file that is read
+rst_prolog = """
+.. |br| raw:: html
+
+   <br />
+"""
+
 # Show or hide TODOs. See http://sphinx-doc.org/ext/todo.html
 todo_include_todos = True
 
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -9,6 +9,12 @@ single system, or you can follow the manual instructions for your OS.
 
 Here's an overview of the options:
 
+* **Vagrant / Virtual Appliance:** Vagrant / OVA is a quick and easy way to try StackStorm.
+  It's already pre-installed, tested and shipped as a virtual image and so saves your time going
+  through time-consuming installation and configuration steps. Works best as a testing,
+  pack development or demo system and recommended to get familiar with StackStorm platform.
+  ``vagrant init stackstorm/st2 && vagrant up`` is all you need to get started.
+  See :doc:`Vagrant </install/vagrant>` for more detailed instructions.
 * **One-line Install:** Run our installation script for an opinionated install of all components
   on a single system. This is a our recommended way to get started. See the :ref:`Quick Install
   <ref-one-line-install>` section below for details.
@@ -21,8 +27,6 @@ Here's an overview of the options:
 * **Ansible Playbooks:** If you are an Ansible user, check these :doc:`/install/ansible` for
   installing |st2|. Ideal for repeatable, consistent, idempotent installation of |st2|.
 * **Docker:** |st2| is now supported on Docker - check out our :doc:`docker` instructions.
-* **Vagrant:** Vagrant is a quick way to spin up a test system - see  `st2vagrant
-  <https://github.com/StackStorm/st2vagrant>`_. This will bring up a new VM, and install |st2|.
 
 Choose the option that best suits your needs.
 
@@ -79,6 +83,7 @@ For more details on reference deployments, or OS-specific installation instructi
 
     Reference Deployment Overview <overview>
     system_requirements
+    Vagrant / OVA <vagrant>
     Ubuntu 14.04 / 16.04 <deb>
     RHEL 7 / CentOS 7 <rhel7>
     RHEL 6 / CentOS 6 <rhel6>
--- a/docs/source/install/vagrant.rst
+++ b/docs/source/install/vagrant.rst
@@ -1,0 +1,115 @@
+Vagrant & Virtual Appliance
+============================
+
+Vagrant / OVA is a quick and easy way to try StackStorm in a VM.<br>
+It's already pre-installed, configured and tested virtual appliance image and so saves time avoiding
+time-consuming installation and configuration steps. Perfect for a quick platform overview,
+testing, demo or even using StackStorm in isolated from the internet air-gapped systems.
+
+We highly recommend to use a Vagrant box to get familiar with the StackStorm platform.
+
+.. contents:: Contents
+   :local:
+
+The source code is available as a GitHub repo:
+`StackStorm/packer-st2 <https://github.com/StackStorm/packer-st2>`_.
+We're welcoming bug reports, feature requests or better pull requests and contributions.
+
+---------------------------
+
+Requirements
+------------
+`Vagrant https://www.vagrantup.com/`_ and `Virtualbox https://www.virtualbox.org/`_ are required.
+If you're not familiar with Vagrant, we recommend to take a look at `Introduction to Vagrant https://www.vagrantup.com/intro/index.html`_.
+
+Quick Start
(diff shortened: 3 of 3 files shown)


docs/source/install/index.rst
```diff
@@ -9,6 +9,12 @@ single system, or you can follow the manual instructions for your OS.
 
 Here's an overview of the options:
 
+* **Vagrant / Virtual Appliance:** Vagrant / OVA is a quick and easy way to try StackStorm.
+  It's already pre-installed, tested and shipped as a virtual image and so saves your time going
+  through time-consuming installation and configuration steps. Works best as a testing,
+  pack development or demo system and recommended to get familiar with StackStorm platform.
+  ``vagrant init stackstorm/st2 && vagrant up`` is all you need to get started.
+  See :doc:`Vagrant </install/vagrant>` for more detailed instructions.
 * **One-line Install:** Run our installation script for an opinionated install of all components
```
armab: I'm thinking about moving the `one-line-install` into separated doc and highlight `vagrant up` way in `index.rst`?

This way we'll push the pre-installed and pre-configured Vagrant box as a way to try & evaluate StackStorm.
This can save some resources in support where people are running `curl|bash` and have issues.

Thoughts?

armab:
@warrenvw This is a first draft. Please take a look.

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. note::
+
+    *Linux login credentials:* |br|
+    Username: ``vagrant`` |br|
+    Password: ``vagrant`` |br|
+
+    *StackStorm login details:* |br|
+    Username: ``st2admin`` |br|
+    Password: ``Ch@ngeMe`` |br|
```
armab: Another idea here is to use `vagrant` as a username and password for everything, incl. stackstorm.

WDYT?

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,121 @@
+.. |br| raw:: html
+
+   <br />
```
armab: @LindsayHill Do you know any other place I can put this block, so we could re-use in all other `.rst` files?

docs/source/install/index.rst
armab: Or alternative, more soft way is to add a new block `Try & Eval` (or similar) before the
https://docs.stackstorm.com/install/index.html#ref-one-line-install `curl|bash` paragraph.

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,121 @@
...
+
+Vagrant & Virtual Appliance
+============================
+Vagrant / OVA is a quick and easy way to try |st2| in a VM.
+
+It's already pre-installed, configured and tested virtual appliance image and so saves time avoiding
+time-consuming installation and configuration steps. Perfect for a quick platform overview,
+testing, demo or even using |st2| in isolated from the internet air-gapped systems.
+
+We highly recommend to use a Vagrant box to get familiar with the |st2| platform.
```
warrenvw: I'd say "We highly recommend using a Vagrant box"...

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,121 @@
...
+
+The source code is available as a GitHub repo:
+`StackStorm/packer-st2 <https://github.com/StackStorm/packer-st2>`_
+
+---------------------------
+
+Requirements
+------------
+`Vagrant <https://www.vagrantup.com/>`_ and `Virtualbox <https://www.virtualbox.org/>`_ are required.
+If you're not familiar with Vagrant, we recommend to take a look at `Introduction to Vagrant <https://www.vagrantup.com/intro/index.html>`_.
```
warrenvw: "we recommend looking at..."

docs/source/install/vagrant.rst
warrenvw: There's a bunch of documentation that refers to the st2admin user and its password. Using `vagrant:vagrant` may increase support costs?

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,121 @@
...
+-------------
+Updating the Vagrant box
+~~~~~~~~~~~~~~~~~~~~~~~~
+Once the newer box version is released, Vagrant will warn you about the available update.
+To update the box:
+
+.. code-block:: bash
+
+    vagrant box outdated
+    vagrant box remove stackstorm/st2
```
warrenvw: Is there any particular reason you're not using `vagrant box update`?

armab: Good catch, let me try that.
It wasn't a thing before.

armab: And so the advantages with `vagrant box remove` Vagrant will ask you to destroy the already running boxes:
```
$ vagrant box remove centos/7
Removing box 'centos/7' (v1803.01) with provider 'virtualbox'...
Box 'centos/7' (v1802.01) with provider 'virtualbox' appears
to still be in use by at least one Vagrant environment. Removing
the box could corrupt the environment. We recommend destroying
these environments first:

default (ID: dddd6bdb40794e41bcc98071b15f8ec7)

Are you sure you want to remove this box? [y/N] y
```

where user has plenty of control.
And this way next `vagrant up` starts with the updated version, instead of running the older one.

warrenvw: Okay, this makes sense. Thanks for the explanation.  👍

docs/source/install/vagrant.rst
LindsayHill: Agreed. Changing username will be a PITA

docs/source/install/index.rst
LindsayHill: You can bump up the priority, but I don't want to have Vagrant over-shadowing everything else.

docs/source/install/vagrant.rst
LindsayHill: There's a `common/` directory in the install docs that I use for other blocks that get included in multiple places. Maybe there?

armab: I was thinking if there is a "maigc file" which is auto-included in all docs, without doing explicitly `include`.

armab: Found it in http://www.sphinx-doc.org/en/stable/config.html#confval-rst_prolog

docs/source/install/vagrant.rst
armab: Makes sense.

I found it's used in a `Quick Start` section: https://docs.stackstorm.com/start.html, - that's important.

The other things are curl|bash + manual install which doesn't apply to Vagrant.

docs/source/conf.py
```diff
@@ -126,6 +126,9 @@ def previous_version(ver):
 .. _exchange: https://exchange.stackstorm.org/
 .. |bwc| replace:: Extreme Workflow Composer
 .. |ipf| replace:: IP Fabric Automation Suite
+.. |br| raw:: html
+
+   <br />
```
LindsayHill: What's the reasoning behind needing this substitution? You shouldn't really need to be inserting HTML line breaks like that.

armab: It's just a common set of shortcuts to re-use.
I think it might be a good instrument to have.

I'm using that `|br|` substitution in several places here: https://github.com/StackStorm/st2docs/blob/60148d18e413e6e15fd60f98974306cdb463fae2/docs/source/install/vagrant.rst

docs/source/install/vagrant.rst
```diff
@@ -0,0 +1,117 @@
...
+
+    Vagrant.configure("2") do |config|
+      config.vm.box = "stackstorm/st2"
+      config.vm.box_version = "2.7.1-20180507"
+    end
+
+The list of available box versions can be found at `Vagrant Cloud <https://app.vagrantup.com/stackstorm/boxes/st2>`_.
+
+
+.. TODO: Add a Vagrant port forwarding example
```
armab: As figured out, we can automatically enable port forwarding rules for a shipped Vagrant box.
I think we should do it by default, make Web UI accessible for example via `localhost:9000` (considering st2 ports are 9001+).
This will simplify user experience a lot.

Created https://github.com/StackStorm/packer-st2/issues/28 to track that.

armab:
Going to merge this soon.

@LindsayHill @warrenvw Any other notes you'd like to share before I'll do it?

docs/source/install/vagrant.rst
LindsayHill: Docs should not be merged until this is figured out. Second thing users will want to do is use Web UI.

docs/source/install/index.rst
```diff
@@ -9,6 +9,12 @@ single system, or you can follow the manual instructions for your OS.
 
 Here's an overview of the options:
 
+* **Vagrant / Virtual Appliance:** Vagrant / OVA is a quick and easy way to try StackStorm.
+  It's already pre-installed, tested and shipped as a virtual image and so saves your time going
+  through time-consuming installation and configuration steps. Works best as a testing,
+  pack development or demo system and recommended to get familiar with StackStorm platform.
+  ``vagrant init stackstorm/st2 && vagrant up`` is all you need to get started.
```
LindsayHill: I don't like having this as the top-level option. We can have "Virtual Appliance" as first option, but I don't like having Vagrant like that, especially when most people don't have Vagrant installed, and are going to use whatever IaaS they are familiar with (local vSphere, AWS, etc)

Maybe we need to separate this into "Quick demo" and "Production" options.

LindsayHill:
My main concern right now is the way we seem to conflate a Vagrant box with a Virtual Appliance. They're related, but they're not the same.

docs/source/install/vagrant.rst
armab: Alright, solved by https://github.com/StackStorm/packer-st2/pull/29.
Here is how it'll work and look like: 

vagrant-up

docs/source/install/index.rst
armab: If we mean OpenSource & DevOps oriented-people, - Vagrant is definitely is used and known.

I really would like to push pre-packaged Vagrant as a first thing here, so people meeting with StackStorm for the first time have easiest, fastest and most reproducible option without going left with other installation methods they don't need yet or not ready to try yet.

I believe this approach will save our time, support efforts and improve first overall StackStorm user experience.

armab:
This PR was in open/RFR status for a plenty time to receive feedback.
Merging.

PR #748, opened 2018-06-19
Added docs for WinRM runners

nmaludy:
Documentation updates for corresponding `st2` PR https://github.com/StackStorm/st2/pull/4169

The build of these docs may fail until the the above PR is merged because the `winrm_*` runners don't exist so their documentation won't be auto-generated.


docs/source/_includes/runner_parameters/winrm_cmd.rst
```diff
@@ -0,0 +1,14 @@
+.. NOTE: This file has been generated automatically, don't manually edit it
```
LindsayHill: Did you manually create this file? It _should_ get generated by https://github.com/StackStorm/st2docs/blob/master/scripts/generate-runner-parameters-documentation.py, called by the Makefile here: https://github.com/StackStorm/st2docs/blob/master/Makefile#L67

Obviously this bit won't work until the runners actually get checked in though

LindsayHill:
I wonder if we should mark the older Windows runners as DEPRECATED in 2.9? Then they can be removed in 3.1 or later. 

Or we might DEPRECATE in 3.0, and remove in 3.2. That gives us an extra cycle (2.9) with the WinRM runners published to a wider audience. 

Maybe also say in the docs "These runners are the future, and we will be deprecating the older Windows runners in the near future."

docs/source/_includes/runner_parameters/winrm_cmd.rst
nmaludy: Yes, i manually cloned down my `st2` fork and checked out my `feature/winrm-runner` branch.

Should i delete this file?

nmaludy:
@LindsayHill I like the idea of deprecating the old runners. I like the idea of deprecating in 3.0 and remove in 3.2. I'll let you guys make the final decision on the timing and versions where this should happen.

I'll add a note about "WinRM is the future" and "windows runners will be deprecated soon".

docs/source/_includes/runner_parameters/winrm_cmd.rst
LindsayHill: Doing a manual clone/branch was fine for testing, but they should not be included in the PR. They _should_ get auto-generated once the upstream changes are in place. For now, that will result in build errors here.

I think it will be 2.9 that we merge those, so we won't merge this docs change until after 2.8 release (which is not far away)

Kami:
Fix invalid syntax (0fbe16dba34dd47c6be914a3323ac023251d6388), moved from copy and paste approach to includes (87e565f5e660c2e2083aa32de095dcf0843072f5) and merged those changes into master.

Thanks!

PR #749, opened 2018-06-20
Upgrade notes for new st2timersengine

lakshmi-kannan:
DO NOT MERGE UNTIL https://github.com/StackStorm/st2/pull/4180 is merged. 

Related: https://github.com/StackStorm/st2-packages/pull/564

Closes https://github.com/StackStorm/st2docs/issues/766


lakshmi-kannan:
Build break is fixed in https://github.com/StackStorm/st2docs/pull/750. Orchestra runner apparently doesn't have runner_parameters.

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,39 @@
 Upgrade Notes
 =============
 
+|st2| v2.9
+----------
+
+* |st2| timers used to be run as part of ``st2rulesengine`` process until versions older than ``v2.9``.
```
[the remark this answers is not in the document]
cognifloyd: Maybe reword the first couple sentences (everything is the same after ` ``st2timersengine`` is the new...`):
```
* |st2| timers moved from the ``st2rulesengine`` to the ``st2timersengine`` in ``v2.9``. Moving timers
  out of the rules engine allows scaling rules and timers independently. ``st2timersengine`` is the new
  process that schedules all the user timers. Please note that when upgrading from older versions, you
  will need to carefully accept changes to ``st2.conf`` file. Otherwise, you risk losing access to
  ``st2`` database in MongoDB.
```

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,39 @@
...
+  If you are already using a ``timer`` section in ``/etc/st2/st2.conf``, you can append the logging
+  configuration parameter to the timer section too.
+
+  .. code-block:: ini
+
+    [timer]
+    local_timezone = America/Los_Angeles
+    logging = conf/logging.timersengine.conf
+
+  Though ``timer`` section in config is supported for backward compatibility, it is recommended to
```
cognifloyd: Possible alternate wording for this section:
```
We recommend renaming the ``timer`` config section to ``timersengine``. Though deprecated, using the
``timer`` section is still supported for backwards compatibility. In a future release, support for
the ``timer`` section will be removed and ``timersengine`` will be the only way to configure timers.
```

lakshmi-kannan:
@cognifloyd bdf456139af1bcd52d0d0b658d31abb12923fddd. I dropped some "the" because it sounded like there is a specific one but in HA there are multiple rules engine. So the definite article seemed unnecessary. LMK what you think.

docs/source/reference/ha.rst
```diff
@@ -79,18 +79,21 @@ sensors is likely to be an enhancement to |st2| in future releases.
...
-release.
+more of the processes becomes unavailable.
+
+st2timersengine
+^^^^^^^^^^^^^^^
+
+``st2timersengine`` is responsible for scheduling all user specified timers. See
+:ref:`timers <ref-rule-timers>` for the specifics on setting up timers via rules.
+
+You have to have one active ``st2timersengine`` process running to schedule all timers. This is trivial to setup in Kubernetes so there is exactly one active container running ``st2timersengine`` process. Failover is handled natively by Kubernetes. In non Kubernetes deployments, external monitoring needs to setup and a new ``st2timersengine`` process needs to be spun up to address failover.
```
armab: In HA doc I don't think we should mention any Kubernetes specifics, but have general descriptions as we do for all other services.

lakshmi-kannan: Yeah, I thought about it but then we need to say how to handle failover. Should we just leave out the k8s part?

lakshmi-kannan: >While Kubernetes can "guarantee something", obviously container/process could be killed by whatever reasons and there is no other timersengine that will keep running.
Won't this be taken care of by Kubernetes though? Isn't that the whole point of using it? 

>This sounds like we don't have any way to promise timers in HA. I mean can't run at least 2+ instances.
Correct, this is what we decided to do with timers. If we decide to solve this, then we need to look at leader election which we intentionally decided to avoid. See https://github.com/StackStorm/discussions/issues/305

>I think it's worth mentioning what happens in scenario when timersengine is not available, if we can't guarantee HA for it. 
There is no A at that point leave alone HA :). It goes without saying IMO but I'll make that explicit. 

>Additionally, what's needed for timersengine to run properly (DB, MQ, anything else), as it's described for other services. Are there any other services that rely on timers functionality?
+1

armab: Pod/container failure will lead to reschedule/restart. What happens in between is a downtime.

Related to that, we'll need to add `/status` for each st2 service https://github.com/StackStorm/k8s-st2/issues/5 so K8s can control the reschedule and knows whether the service is really alive or just sits there in any manner of deadlock, spinning the cycles and being actually non-responsive. See https://github.com/StackStorm/st2/issues/4020


With no big A in HA, we still can haz High :smiley:

docs/source/reference/ha.rst
```diff
@@ -79,18 +79,28 @@ sensors is likely to be an enhancement to |st2| in future releases.
...
+st2timersengine
+^^^^^^^^^^^^^^^
+
+``st2timersengine`` is responsible for scheduling all user specified timers. See
+:ref:`timers <ref-rule-timers>` for the specifics on setting up timers via rules.
+``st2timersengine`` process needs access to both Mongo database and RabbitMQ message bus.
+
+You have to have exactly one active ``st2timersengine`` process running to schedule all timers.
+Having more than one active ``st2timersengine`` will result in duplicate timer events and therefore
+duplicate rule evaluations leading to duplicate workflows or actions.
```
armab: Very important detail here was just documented :+1:

armab:
@Kami I guess you wanted to merge, but closed instead?

PR #750, opened 2018-06-20
Not all runners (e.g. orchestra) have runner_parameters

lakshmi-kannan:


scripts/generate-runner-parameters-documentation.py
```diff
@@ -66,19 +66,21 @@ def main():
...
-            result.append(line)
-
-        file_name = runner['name'].replace('-', '_')
-        path = '../docs/source/_includes/runner_parameters/%s.rst' % (file_name)
-        destination_path = os.path.join(CURRENT_DIR, path)
-        result = '\n'.join(result)
-
-        with open(destination_path, 'w') as fp:
-            fp.write(result)
+        runner_parameters = runner.get('runner_parameters', None)
```
Kami: LGTM, although I would prefer `runner.get('runner_parameters', {})`, then we can get rid of `if` and one level of indentation :P

armab:
Would be better to communicate in PR description more nicely what is it and why it's needed.

I dig through links that this PR fixes the st2docs build, but others might guess on our cryptic changes.

scripts/generate-runner-parameters-documentation.py
lakshmi-kannan: Won't work in this case since the file generation and writing etc is needed only when runner_parameters are present but otherwise your suggestion is valid.

PR #753, opened 2018-06-25
Add docs section for the new Orchestra workflow engine

m4dcoder:
Add docs section for the new Orchestra workflow engine. Add script to clone the docs from the Orchestra repo at https://github.com/StackStorm/orchestra and copy the relevant files locally. Run the clone script on make docs and make livedocs.


.gitignore
```diff
@@ -17,3 +17,11 @@ docs/source/_static/images/ztp_topology.jpg
...
 docs/source/_static/images/ztp_usecase_2.jpg
 .vscode
+
+# Orchestra Copies
+docs/source/orchestra/overview.rst
+docs/source/orchestra/expressions.rst
+docs/source/orchestra/context.rst
+docs/source/orchestra/yaql.rst
+docs/source/orchestra/jinja.rst
+docs/source/orchestra/languages
```
LindsayHill: Wonder if we should just do `docs/source/orchestra` directory here, rather than specific files?

scripts/clone-orchestra.sh
```diff
@@ -0,0 +1,21 @@
...
+fi
+
+# Only supports master branch for duration of beta.
+REQUIRED_BRANCH="master"
+
+echo "Cloning branch $REQUIRED_BRANCH of orchestra..."
+git clone -b $REQUIRED_BRANCH https://github.com/StackStorm/orchestra.git ${REPO_DIR}
+
+echo "Copying files to local st2docs repo..."
+cp ${REPO_DIR}/docs/source/overview.rst ./docs/source/orchestra/overview.rst
```
Kami: I would prefer to use a git subtree over this approach.

It's confusing and also causes build issues.

PR #758, opened 2018-06-29
And more info about the timezone format for core.st2.CronTimer trigger

longfei-zhang:
Add more info for the timezone parameter of core.st2.CronTimer trigger,
the timezones use the pytz format, e.g. `Asia/Shanghai`.


Kami:
LGTM, thanks :+1:

PR #760, opened 2018-07-02
Add upgrade notes section for v2.8 release

Kami:
New ``st2workflowengine`` service doesn't start automatically on upgrade due to ``st2.conf`` config file missing ``workflow_engine.logging`` section.

Keep in mind that this section doesn't get automatically added on upgrade if user chooses to keep the existing ``st2.conf`` file and doesn't manually resolve the conflicts between the existing and new config file (that's actually desired, because if a user would select to pick a new st2.conf file, it would wipe out the existing settings and credentials and manual merge / diff is also error prone).


Kami:
/cc @m4dcoder @bigmstone 

@lakshmi-kannan You will likely encounter the same issue with new timer service in v2.9.0.

PR #769, opened 2018-07-24
Added puppet-st2 documentation

nmaludy:
`puppet-st2` just reached the `1.0` milestone and wanted to document it as an installation option in the official StackStorm docs.

This document will grow over time as more features and enhancements are added to the `puppet-st2` module.

### DIFF
--- a/docs/source/install/puppet.rst
+++ b/docs/source/install/puppet.rst
@@ -1,11 +1,14 @@
 Puppet Module
 =============
 
-If you're ready to take complete control of your |st2| instances, then the ``stackstorm-st2``
-is for you! It offers repeatable, configurable, and idempotent production-friendly
-|st2| installations.
+If you're ready to take complete control of your |st2| instances, then ``stackstorm-st2``
+Puppet module is for you! It offers repeatable, configurable, and idempotent
+production-friendly |st2| installations.
 
-The source code for all the Puppet module is available as a GitHub repo:
+The ``stackstorm-st2`` Puppet Module is available on Puppet Forge:
+`stackstorm-st2 <https://forge.puppet.com/stackstorm/st2>`_
+
+Source code for the module is available as a GitHub repo:
 `StackStorm/puppet-st2 <https://github.com/stackstorm/puppet-st2/>`_
 
 .. contents:: Contents
--- a/docs/source/install/puppet.rst
+++ b/docs/source/install/puppet.rst
@@ -29,7 +29,7 @@ Quick Start
 -----------
 
 The first step is installing Puppet, for this please consule the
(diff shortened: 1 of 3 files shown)


docs/source/install/puppet.rst
```diff
@@ -0,0 +1,310 @@
...
+The first step is installing Puppet, for this please consule the
+`Puppet Installation documentation <https://puppet.com/docs/puppet/latest/install_linux.html>`_
+
+To get started with a single node deployment, and default configuration settings, run these
+commands (expected to run as `root`):
+
+.. code-block:: bash
+
+  mkdir modules/
+  git clone https://github.com/StackStorm/puppet-st2 modules/st2
```
armab: Is it possible to rely on instructions with "official" Puppet forge, instead of going the git clone way (which relies on latest "development" code)?

nmaludy:
@armab great idea, updated the docs!

docs/source/install/puppet.rst
```diff
@@ -0,0 +1,305 @@
...
+commands (expected to run as `root`):
+
+.. note::
+   
+  Puppet versions <= 3.x are no longer supported. Please utilize Puppet >= 4.
+
+.. code-block:: bash
+
+  # create a Puppetfile that defines our StackStorm module
+  echo -e "forge 'https://forgeapi.puppetlabs.com'\n mod 'stackstorm/st2'" > Puppetfile
```
armab: Instead of scripting, it would be nicer to extract this part to just highlight the resulting `Puppetfile` user needs to create via code block: 
```
https://forgeapi.puppetlabs.com
mod 'stackstorm/st2'
```

Hopefully would make it a bit more readable.

docs/source/install/puppet.rst
```diff
@@ -0,0 +1,311 @@
...
+  Puppet versions <= 3.x are no longer supported. Please utilize Puppet >= 4.
+
+First, create a file called ``Puppetfile`` with the following content:
+
+.. code-block:: ruby
+
+  forge 'https://forgeapi.puppetlabs.com'
+  mod 'stackstorm/st2'
+
+Next, utilize ``librarian-puppet`` to download and install the ``puppet-st2`` module
```
armab: Wondering if we could minimize this part, including manual instructions with librarian, which looks a bit hacky.

Look at https://github.com/StackStorm/st2docs/blob/f5658ffc3e8ee4846d9faec77dbed182c538f663/docs/source/install/puppet_chef_salt_ansible.rst#manual-installation if we could take examples of the previous Puppet installation instructions.

For example, is 
```
puppet module install stackstorm-st2
```
still a possible thing?

armab: Since I'm not familiar with Puppet, curious if `Quick Start` could be really quick :)

PR #771, opened 2018-07-27
Explicitly install crudini (it's used later on)

blag:
It's used in the "Setup Datastore Encryption" section (`docs/source/install/common/datastore_crypto_key.rst`) but it's never installed anywhere.

Don't know what the package is called on (or what repos to use) for RHEL systems. 😕

### DIFF
--- a/docs/source/install/deb.rst
+++ b/docs/source/install/deb.rst
@@ -34,6 +34,7 @@ Install MongoDB, RabbitMQ, and PostgreSQL:
   EOT"
   sudo apt-get update
 
+  sudo apt-get install -y crudini
   sudo apt-get install -y mongodb-org
   sudo apt-get install -y rabbitmq-server
   sudo apt-get install -y postgresql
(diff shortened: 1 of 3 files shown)


nmaludy:
```
Name        : crudini
Arch        : noarch
Version     : 0.9
Release     : 1.el7
Size        : 60 k
Repo        : installed
From repo   : epel
Summary     : A utility for manipulating ini files
URL         : https://github.com/pixelb/crudini
License     : GPLv2
Description : A utility for easily handling ini files from the command line and shell
            : scripts.
```

Make sure EPEL repo is setup (to do this you install `epel-release` package)

blag:
@nmaludy Thanks! Looks like those repos are already activated/installed for other packages, so I just need to `yum install`.

LindsayHill:
I think that the builds are failing because it's cached something that includes older orchestra references that have not been fully updated to orquesta. Should be OK if we just merge this here.

PR #772, opened 2018-07-27
Typo in remote URL

LindsayHill:


LindsayHill:
Bit embarrassing we missed it earlier, since that line is specifically calling out that different URL

PR #774, opened 2018-07-30
Add backslash for escape sequences

pengyao:
`$releasever` will be null character without escape character in bash


LindsayHill:
Thanks, change looks good. Build failures are due to upstream repo/package name changes. Will merge this once those get fixed.

LindsayHill:
Replaced by #776, for some reason builds kept failing here.

PR #780, opened 2018-08-10
Use backticks around dashes and slashes

blag:
...because ReStructured Text renders double dashes into a single wide dash.

See the `kwarg_op` bullet point at https://docs.stackstorm.com/reference/runners.html#id1


blag:
CI seems to be failing on Orchestra -> Orquesta rename (unrelated to these changes).

LindsayHill:
Those files are auto generated. You need to make changes in those scripts, rather than the output

LindsayHill:
I think this is the script that you need to update https://github.com/StackStorm/st2docs/blob/master/scripts/generate-runner-parameters-documentation.py

or https://github.com/StackStorm/st2docs/blob/master/scripts/generate-overridable-runner-parameter-attributes-list.py

LindsayHill:
And those CI failures are because your branch is 43 commits behind master. Update your branch.

PR #784, opened 2018-08-13
Explicitly create a Python2 virtualenv

vutny:
Greetings here!

In modern Debian and Ubuntu distributions, the `virtualenv` package installed from repositories creates Python3 virtual environment by default. Since the docs are still building with Py2, we just need to provide a path to correct binary executable. It also should be safe for any other platform.

@Kami @LindsayHill Please review. Many thanks!


LindsayHill:
Looks like this really needs to be `--python=python2.7` to work on more OSes (e.g. macOS).

PR #793, opened 2018-09-09
Added note on mistral vs action timeout

LindsayHill:
Trying to reduce confusion between Mistral's `timeout:` parameter, and underlying action `timeout:`.


LindsayHill:
Also needed to update CircleCI configuration - looks like there's been a change in some of our pip dependencies, and now need `python-dev` installed.

PR #801, opened 2018-09-18
Document --single-sensor-mode st2sensorcontainer service CLI option

Kami:
This pull request documents new ``--single-sensor-mode`` sensor container CLI option.

(Will cherry pick it in v2.9 branch once it's merged)

### DIFF
--- a/docs/source/reference/ha.rst
+++ b/docs/source/reference/ha.rst
@@ -77,6 +77,22 @@ implemented with HA in mind so that the same sensor can be deployed on multiple 
 sensor managing active-active or active-passive. Providing some platform level HA support for
 sensors is likely to be an enhancement to |st2| in future releases.
 
+By default sensor container service runs in managed mode. This means that the sensor container
+process manages child processes for all the running containers and restarts them if they crash or
+similar.
+
+In some scenarios this is not desired and service / process life-cycle (restarting, scaling out,
+etc.) is handled by a third party service such as Kubernetes.
+
+To account for such deployments, sensor container can be started in single sensor mode using
+``--single-sensor-mode`` and ``--sensor-ref`` command line options. When those options are
+provided, sensor container service will run a single sensor and exit immediately if a sensor
+crashes or similar.
+
+.. code-block:: bash
+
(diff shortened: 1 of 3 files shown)


armab:
I think this documentation might be a better fit for "Sensors" or "Sensors Partitioning" document and just provide a reference to that doc from HA.

Kami:
@armab Yeah, I was wondering where should I put it - either HA or sensor partitioning section.

@LindsayHill what do you think?

LindsayHill:
Good point @armab - at a minimum it needs cross-linking.

I think it makes sense to have a mention of it in HA docs, and then config details in the sensor partitioning doc.

Kami:
I had a look at sensor partitioning docs again and I don't think it belongs there.

Sensor partitioning and this flag are two unrelated functionalities. Operators who will use ``--single-sensor-mode`` flag will likely (need to) handle sensor partitioning themselves (they will need to manually assign which sensor runs on which node as per discussion with @armab in the past) and not utilize the build in partitioning functionality (in fact, those two things are mutually exclusive).

Similar story for general "Sensors" section - this one is geared towards end users and not operators.

For now, I will keep it in HA document, feel free to update it if you think it should go somewhere else.

Kami:
Added a note / cross-reference in 3e6614b.

docs/source/reference/ha.rst
```diff
@@ -77,6 +77,26 @@ implemented with HA in mind so that the same sensor can be deployed on multiple
...
+To account for such deployments, sensor container can be started in single sensor mode using
+``--single-sensor-mode`` and ``--sensor-ref`` command line options. When those options are
+provided, sensor container service will run a single sensor and exit immediately if a sensor
+crashes or similar.
+
+For example:
+
+.. code-block:: bash
+
+  st2sensorcontainer --single-sensor-mode --sensor-ref linux.FileWatchSensor
```
armab: I think this single-sensor-per-process mode represents manual way of partitioning sensors.

Pretty similar to File-based partitioning https://docs.stackstorm.com/reference/sensor_partitioning.html#file where user defines associations on which node to start this or that sensor.

From this point of view I think it makes sense to move this configuration example from HA to Sensor Partitioning doc.

PR #802, opened 2018-09-19
Small tweaks

blag:
Use HTTPS when importing the nginx signing key, and for the nginx repo `baseurl`.

Also, use actual major version numbers (eg: `6` or `7`) instead of `$releasever`, not only because it's more explicit, but also because it's 90% more efficient to type.


LindsayHill:
This should have accompanying changes in github.com/StackStorm/st2-packages

docs/source/install/deb.rst
```diff
@@ -131,9 +131,9 @@ you will need to add the official Nginx repository:
 .. code-block:: bash
 
   # Add key and repo for the latest stable nginx
-  sudo apt-key adv --fetch-keys http://nginx.org/keys/nginx_signing.key
+  sudo apt-key adv --fetch-keys https://nginx.org/keys/nginx_signing.key
```
armab: Were these changes tested end-to-end?

I mean, did we ran this command on a clean Ubuntu14 server first?

Same applies for other deb-based commands to be tested on Ubuntu14 and rpm-based commands to be tested on CentOS6.

blag: Good catch - `apt-key` in Ubuntu 14.04 doesn't have a handler for `https` (🙄), so I removed that commit.

PR #807, opened 2018-09-25
Update notes entry for new dynamic runner loading

Kami:
Upgrade notes entry for change in https://github.com/StackStorm/st2/pull/4217.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -8,6 +8,15 @@ Upgrade Notes
 |st2| v3.0
 ----------
 
+* Old runner names which have been deprecated in |st2| v0.9.0 have been removed. If you still have
+  any actions which refer to runners using old names you need to update them to keep them working.
+
+  * ``run-local`` -> ``local-shell-cmd``
+  * ``run-local-script`` -> ``local-shell-script``
+  * ``run-remote`` -> ``remote-shell-cmd``
+  * ``run-remote-script`` -> ``remote-shell-script``
+  * ``run-python`` -> ``python-script``
+  * ``run-http`` -> ``http-request``
 * In |st2| v2.7 action runner modules have been refactored so they are now fully standalone and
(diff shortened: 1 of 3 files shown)


docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,39 @@
...
+
+  In this release we updated our runner loading mechanism which makes ``/opt/stackstorm/runners``
+  directory obsolete.
+
+  All the runners are now installed as Python packages into |st2| virtual environment
+  (``/opt/stackstorm/st2``) during package build process and dynamically loaded when requested.
+
+  This provides for more flexible installation and loading of runner modules. To install a custom
+  runner, user now just needs to install Python package which contains runner module into |st2|
+  virtual environment and run ``sudo st2ctl reload --register-runners`` command.
```
armab: Just to be sure I'm understanding it right, default runners that we package in core don't need additional `--register-runners` and would work by default on app init?

Kami: This command still needs to be run so the information from disk is stored in the database (same as for any other resources).

In the future though, we could make that step implicit and happen automatically on service start up (like it used to be the case).

armab: In this case, it sounds like a regression to me :-1: 

While ideally to go towards pre-init all the content that comes with st2 core on app startup, making the st2 self-sufficient and minimally working after the first installation (and that really affects all deployments), we sadly go the opposite side offloading repetitive instructions on user's side adding more complexity.

Kami: It's not a regression - it's exactly the same end user behavior, just the internals change.

While I would prefer if we could get rid of it, we need to approach it in incremental steps.

Kami: Having said that, I will see what we can do.

That st2 PR is already growing a bit big and I would like to scope things down.

PR #815, opened 2018-10-29
Note suggesting using yum exclude feature

csv7:
Suggest using exclude=nginx in epel.repo rather than simply disabling epel repo during nginx install. Yum is a vile beast, and sometimes its streams will cross unless explicitly told not to use a repo for a particular package.

### DIFF
--- a/docs/source/install/rhel6.rst
+++ b/docs/source/install/rhel6.rst
@@ -200,7 +200,12 @@ you will need to add the official Nginx repository:
 
   # Install nginx, avoid epel repo in favor of nginx.org
   sudo yum --disablerepo='epel' install -y nginx
-
+  
+  # A better option may be to disable nginx in /etc/yum.repos.d/epel.repo
+  # This sets the exclude flag on all enabled sections of epel.repo, 
+  # and ensures yum won't try to install or update it from epel:
+  sed -i 's/^\(enabled=1\)$/exclude=nginx\n\1/g' /etc/yum.repos.d/epel.repo
+  
   # Install st2web
   sudo yum install -y st2web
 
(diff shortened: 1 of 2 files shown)


armab:
@LindsayHill WDYT? Is the instructions there are more about being prod-focused or for demonstration/try purposes for users to understand the process?

For first "prod" case, I think we should include it, for second "demo" and simplicity reasons, - the change would be a bit excessive.

LindsayHill:
I'm inclined to take it a step further. Rather than hedge, with a suggestion, instead replace the original method with the new one. So instead of:

```
  # Install nginx, avoid epel repo in favor of nginx.org
   sudo yum --disablerepo='epel' install -y nginx

   
   # A better option may be to disable nginx in /etc/yum.repos.d/epel.repo
   # This sets the exclude flag on all enabled sections of epel.repo, 
   # and ensures yum won't try to install or update it from epel:
   sed -i 's/^\(enabled=1\)$/exclude=nginx\n\1/g' /etc/yum.repos.d/epel.repo
```

Instead just say:
```
   # Ensure that EPEL repo is not used for nginx:
   sed -i 's/^\(enabled=1\)$/exclude=nginx\n\1/g' /etc/yum.repos.d/epel.repo

   # Install nginx
   sudo yum install -y nginx
```

**and** update the bootstrap scripts for [EL6](https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el7.template.sh#L299) and [EL7](https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el7.template.sh#L300)

blag:
Closing since this was fixed in #838 and StackStorm/st2-packages#602.

PR #816, opened 2018-10-29
use repo exclude feature

csv7:


armab:
Duplicate of #815

LindsayHill:
As per note, this is duplicate. (think this happened during the GitHub outage)

PR #835, opened 2018-12-19
Pin to < 1.8 for Sphinx

LindsayHill:
Having problems with search with Sphinx 1.8.x, so will pin to 1.7.x for now.


LindsayHill:
Related: https://github.com/rtfd/sphinx_rtd_theme/pull/672

We need to merge that change into our theme fork

Issue #847, opened 2019-02-06
Document using SSL / TLS for message bus (RabbitMQ) connection


Kami:
Right now we don't have any docs how to use SSL / TLS for RabbitMQ authentication and connection (we have docs for that for MongoDB, but not for RabbitMQ).

We should add docs for it and also verify it works with the current code base (might need small changes in the docs to support it fully).

Kami:
I confirmed using SSL connection works our of box. I will document that.

Supporting client side certificate auth will require some small code changes.

Kami:
Resolves via #848, #850.

PR #848, opened 2019-02-06
Document how to use TLS/SSL with RabbitMQ

Kami:
This pull request documents how to enable TLS for RabbitMQ connections.

Part of #847.

### DIFF
--- a/docs/source/install/config/config.rst
+++ b/docs/source/install/config/config.rst
@@ -109,7 +109,7 @@ For example:
 
 Keep in mind that RabbitMQ server needs to be configured to enable TLS support.
 For more information, refer to the official documentation - `Enabling TLS Support in RabbitMQ
-<https://www.rabbitmq.com/ssl.html#enabling-tls>`).
+<https://www.rabbitmq.com/ssl.html#enabling-tls>`_.
 
 .. _ref-rabbitmq-cluster-config:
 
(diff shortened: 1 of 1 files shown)


docs/source/install/config/config.rst
```diff
@@ -98,6 +98,19 @@ In :github_st2:`/etc/st2/st2.conf <conf/st2.prod.conf>` include the following se
...
 The ``#RMQ_VHOST`` property is optional and can be left blank.
 
+|st2| also supports SSL/TLS to encrypt RabbitMQ connections. To enable SSL, you need to pass
+``?ssl=true`` query parameter at the end of the connection URL.
+
+For example:
+
+.. code-block:: ini
+
+   url = amqp://guest:guest@127.0.0.1:5671/?ssl=true
```
armab: Is there anything else that's required on ST2 conf side to configure certificates/trusted/chain to communicate with SSL-enabled RabbitMQ server, the right way?

Just for reference, I'm looking at RabbitMQ clients from the examples like:
https://www.rabbitmq.com/ssl.html#java-client-connecting-with-peer-verification
https://pika.readthedocs.io/en/latest/examples/tls_mutual_authentication.html
http://docs.celeryproject.org/projects/kombu/en/latest/reference/kombu.connection.html

Kami: This change only enables secure communication to the server using TLS, but it doesn't do peer verification and it also doesn't support client certificate based authentication.

For that we will need to modify the code (similar to the user contribution for MongoDB not too long ago).

The plan is to use the same configuration option names as we use for MongoDB in the ``database`` section in st2.conf (for consistency reasons).

But yeah, that will be a separate change. I will open an issue in StackStorm/st2 to track it.

armab: Would be good to add a small note about that part in st2docs then.

Kami:
I was actually also hoping to implement those missing config options before v2.10.2 will be released, but I will add a note about that and merge it for now.

PR #849, opened 2019-02-08
Push built docs direct from CircleCI to S3

LindsayHill:
@Kami here's an initial pass at changing our Docs build process.

Our CircleCI checks here already build the docs as part of testing them, so we should just re-use that, rather than doing yet another rebuild on our CICD systems.

Store the docs in the workspace, then on commits to master or vX.Y, restore from that workspace and push the built docs to S3.

Note that current process copies pre-release docs to `docs.stackstorm.com/latest` as well as `/docs.stackstorm.com/3.0dev`. I don't think there's any value in copying content to the `3.0dev` directory, for pre-release versions. There *is* value in copying GA releases to their own folder, i.e. `/2.8`, `/2.9`, etc, 

TODO:

- [x] Figure out how to programmatically identify the current GA version, and if our branch matches that, then *also* copy docs to ${BUCKETNAME}/ (that is, the top level of docs.stackstorm.com). ~Would appreciate some help on that~ - NOW DONE

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,19 @@ jobs:
               aws s3 sync generated-site/st2docs/* \
               s3://${EWCDOCS_BUCKET}/${CIRCLE_BRANCH} --dryrun
             fi
+      - run:
+          # Check the install scripts to see if this is the current GA version
+          name: Check if current GA branch, in which case also deploy to main site
+          command: |
+            GA_VER=$(curl -sSL https://stackstorm.com/packages/install.sh|grep ^BRANCH=|sed 's/[^0-9.]*//g')
+            if [ "${CIRCLE_BRANCH}" = "${GA_VER}" ]; then
+              aws s3 sync generated-site/st2docs/* \
+              s3://${ST2DOCS_BUCKET}/ --dryrun
+              aws s3 sync generated-site/ewcdocs/* \
+              s3://${EWCDOCS_BUCKET}/ --dryrun
+            else
+              echo "Not current GA version"
+            fi
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       - run: make docs
       - run:
           name: Store HTML docs in workspace dir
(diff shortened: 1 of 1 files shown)


LindsayHill:
@Kami this is now RFR. I think it is all working OK. Just before merging it, we need to disable the web hook that triggers the current build & push process.

I'll also submit a similar PR for github.com/StackStorm/ipfabric-docs

.circleci/config.yml
```diff
@@ -30,13 +39,63 @@ jobs:
...
       - ewcdocs
+      - deploy:
+          requires:
+            - st2docs
+            - ewcdocs
+          filters:
+            branches:
+              only:
+                - master
+                - /v[0-9]+\.[0-9]+/
```
armab: I guess successful deployment will be a best reviewer.
You can temporary add `docs_build` branch to the allowed list to verify if deployment really works as expected.

.circleci/config.yml
```diff
@@ -13,6 +15,13 @@ jobs:
           key: v1-dependency-cache-{{ checksum "st2/requirements.txt" }}
       - run: sudo apt install python-dev
       - run: make docs
+      - run:
+          name: Store HTML docs in workspace dir
+          command: mkdir /tmp/st2docs && cp -r docs/build/html/* /tmp/st2docs/
+      - persist_to_workspace:
```
armab: Does it also makes sense to copy those dirs to CircleCI artifacts?
May help with some future debugging.

.circleci/config.yml
LindsayHill: Yep, I tried that - I've removed it from the final PR, but I did that during my tests. Worked as expected (once I fixed a couple of bugs :))

.circleci/config.yml
LindsayHill: I wondered about that. Might make sense.

LindsayHill: Done - now storing artifacts. Works OK. 
https://github.com/StackStorm/st2docs/pull/849/commits/5aa1246438fff60a9fbdde27342b16ce34a519d8

PR #859, opened 2019-03-04
Added a foot note about the necessity to specify dependent libraries with specifying version in the requirements.txt.

userlocalhost:
This change corresponds to the change of StackStorm/st2#4571.


Kami:
Merged, thanks.

PR #864, opened 2019-03-08
Change group on examples pack

LindsayHill:


bmomberger-bitovi:
I don't know if this is everyone's experience, but for me the examples/ dir gets copied with 755 permissions, while the preinstalled packs have 775 (which lets the st2packs group write to it).

So probably also need `chmod -R g+w /opt/stackstorm/packs/examples`

LindsayHill:
Could well be that. 

Interesting thing is that every ST2 RPM/Deb update also goes & resets all those permissions, fixing them up so group is `st2packs`, and has g+w.

LindsayHill:
I'll update the docs

PR #867, opened 2019-03-13
Update pack documentation

yaleman:
Pack documentation doesn't make mention of the fact that local directories that are git repos only install the latest commit, not the latest file edit.

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -102,7 +102,7 @@ Exchange, and you can install your own packs from git just as easily.
     # Install your own pack from git
     st2 pack install https://github.com/emedvedev/chatops_tutorial
 
-By default, the latest version of a pack will be installed, but you can specify a particular
+By default, the latest commit to a pack repository will be installed, but you can specify a particular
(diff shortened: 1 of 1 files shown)


docs/source/packs.rst
```diff
@@ -123,6 +123,9 @@ Finally, you can install a pack from existing local dir:
     # Install a pack from '/home/stanley/bitcoin' dir
     st2 pack install file:///home/stanley/bitcoin
 
+.. warning::
+
+    Installing a pack from a directory that is a git repository will only install the latest commit, and ignores any subsequent changes to the files.
```
Kami: Perhaps we should make it more explicit and say something like ``and ignores any subsequent uncommitted changes to the files.``?

Since ``ignores any subsequent changes to the files`` by itself is a bit confusing.

Kami:
Thanks.

I will make a small modification (add ``uncommitted`` to make it more explicit) and merge it into master.

Kami:
Merged, thanks.

PR #871, opened 2019-03-28
Add upgrade note for new config option

Kami:
This pull request adds an upgrade note for new st2.conf config option.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -49,6 +49,9 @@ After:
     enable = True
     backend = enterprise
 
+After you do that, you need to restart st2api service for changes to take affect- ``sudo st2ctl
+restart-service st2api``.
+
 .. _ref-upgrade-notes-v2-10:
 
 |st2| v2.10
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -3,7 +3,7 @@
 Upgrade Notes
 =============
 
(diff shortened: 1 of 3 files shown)


docs/source/upgrade_notes.rst
```diff
@@ -28,8 +28,40 @@ Upgrade Notes
    /opt/stackstorm/st2/bin/pip install "git+https://github.com/StackStorm/stackstorm-runner-windows.git#egg=stackstorm-runner-windows"
 
    sudo st2ctl reload --register-runners
- 
- * The :doc:`Inquiries </inquiries>` API has been promoted from the ``/api/exp`` path to ``/api/v1``. If you have any external systems that use this API they will need to be updated to use the new path. st2client has been updated to use the new path.
+* The :doc:`Inquiries </inquiries>` API has been promoted from the ``/api/exp`` path to ``/api/v1``.
+  If you have any external systems that use this API they will need to be updated to use the new
+  path. st2client has been updated to use the new path.
+* If you are using |EWC| with RBAC you need to update your ``/etc/st2/st2.conf`` config file for RBAC
+  to work after the upgrade.
```
armab: I thought this migration should be done by the corresponding package we add & install as a dependency?

Kami: It does happen automatically now inside post install, yes (everything except restarting the services).

I left it here just in case (to ensure option is indeed there) and also to serve as a "reference" when this option was added.

PR #872, opened 2019-03-29
Add configuration steps for Microsoft Teams to ChatOps

blag:
Cleaned up the `Makefile` a tiny bit.


docs/source/chatops/aliases.rst
```diff
@@ -369,6 +369,59 @@ dynamically:
 Support for other chat providers is coming soon, and of course you are always welcome to
 contribute! See the example below for hacking on ``extra``.
 
+Mattermost and Rocketchat also support the Slack attachments API. However, you will need
+to use the ``mattermost`` and ``rocketchat`` keys to ``extra``:
```
Kami: "you will need to add .... to extra"?

docs/source/chatops/index.rst
```diff
@@ -50,6 +50,7 @@ Interested in learning more? Here are some things to get you started on your voy
     :maxdepth: 1
 
     chatops
+    Configuration for Microsoft Teams <msteams_is_a_diva>
```
Kami: Perhaps we should have an "Adapters" section where we list supported adapters, their configuration option and limitations?

docs/source/chatops/msteams_is_a_diva.rst
```diff
@@ -0,0 +1,234 @@
...
+      and send events via a websocket. We have investigated this method but we were unable
+      to get it working in development.
+
+Setup BotFramework
+~~~~~~~~~~~~~~~~~~
+
+1. Note the internet accessible hostname and port for your st2chatops (hubot) server.
+2. Register a bot within `BotFramework <https://dev.botframework.com/bots/new>`_.
+3. Note your Microsoft App ID and password, you will need these later.
+4. Configure the "Messaging endpoint" option to your st2chatops server, and append
```
Kami: Some screenshots wouldn't hurt to make it even more "dummy" proof.

docs/source/chatops/aliases.rst
blag: Fixed my grammar to be more clear about what I'm pointing out.

docs/source/chatops/msteams_is_a_diva.rst
blag: Good idea, but I'll handle that later. Right now, even though it's officially supported, I don't consider the adapter 100% GA, more like beta quality. Once it gets to GA status I'll circle back to this.

Although hopefully we have Orky working by then.

blag:
Running tests locally before I push up some more changes.

docs/source/chatops/index.rst
blag: Excellent idea - done!

blag: Well, somewhat done. I don't list the exact options and limitations, but I do point people to the source code so they can figure out what to do themselves. Since this is more of an advanced option, I'm not too worried about not having perfect documentation for it.

docs/source/chatops/msteams_is_a_diva.rst
```diff
@@ -0,0 +1,234 @@
+Using Microsoft Teams (with BotFramework)
```
LindsayHill: I strongly recommend not using a page name like that. At some point that comment might not apply, and changing page names changes links, which means you'll need to add redirect pages or Google will point to the old content forever

blag: Fixed.

armab:
Indeed very involved configuration for MSTeams :disappointed: 
Requirement to expose an HTTP(s) endpoint to public is even more daunting. Good to know this for K8s context.

PR #886, opened 2019-04-18
Add a note on using "secret: true" for action parameter default values which use decrypt_kv Jinja filter

Kami:
This comes from my comment / find here - https://github.com/StackStorm/st2/pull/4640#discussion_r276593698

Per the comment there - not using ``secret: true`` on default action parameter values which use ``decrypt_kv`` Jinja filter is almost never a good idea. If such parameter is not marked as a secret, every user who has permission to run that action will be able to see raw unencrypted value when running the action.

This pull request documents that potential security issues.

As an additional improvement (and I do think we should do something along those lines and provide a more safe and reasonable default behavior), we have a couple of options:

1. Don't allow using ``decrypt_kv`` on non-secret parameters. Imo, that's quite reasonable.
2. Default ``secret: True`` for action parameters which use ``decrypt_kv`` filter on a default value. This way it would be an opt-in and user would need to explicitly mark it as ``secret: false``. Downside of this approach is that it's less explicit.
3. Warn user when running ``st2-register-content`` if an action parameter default value uses ``decrypt_kv`` filter, but it's not marked as a secret. That's potentially in addition to 2).


Kami:
@userlocalhost^ just a heads up - your feedback on possible solutions for a more reasonable and safe default behavior is also welcome.

punkrokk:
@Kami / @armab any reason not to update this for merging. It's a useful update

docs/source/datastore.rst
```diff
@@ -625,16 +626,22 @@ only. To get plain text, please run the command with the ``--decrypt`` flag:
     able to decrypt every value) and by the user who set that value in case of the user-scoped
     datastore items (i.e. if ``--scope=user`` flag was passed when originally setting the value).
 
-If you are using system scoped variables (``st2kv.system``) to store secrets, you can decrypt them
-and use as parameter values in rules or actions. This is supported via Jinja filter ``decrypt_kv``
-(read more about :ref:`Jinja filters<applying-filters-with-jinja>`). For example,
-to pass a decrypted password as a parameter, use:
+If you are using system scoped (``st2kv.system``) or user scoped (``st2kv.user`` )datastore items
```
armab:
```suggestion
If you are using system scoped (``st2kv.system``) or user scoped (``st2kv.user``) datastore items
```

armab:
Yeah, looks indeed useful.
@punkrokk Thanks for noticing this and closing the loops, please feel free to merge it!

PR #888, opened 2019-04-18
Remove beta tag from Inquiries

LindsayHill:
Remove beta tag, change Mistral references in Inquiries docs to Orquesta

### DIFF
--- a/docs/source/inquiries.rst
+++ b/docs/source/inquiries.rst
@@ -1,28 +1,19 @@
 Inquiries
 =========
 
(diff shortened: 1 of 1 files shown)


docs/source/inquiries.rst
```diff
@@ -1,28 +1,19 @@
 Inquiries
 =========
 
-|st2| 2.5 introduced a new feature that allows you to pause a workflow to wait for additional
```
armab: ... that was quite a long beta :)

PR #889, opened 2019-04-24
Migrate to Sphinx2 and Python3

LindsayHill:


LindsayHill:
@Kami this is now working, and builds docs using Sphinx2 + Python3. The published artifacts look OK - e.g. see here https://1870-49105003-gh.circle-artifacts.com/0/st2docs/index.html

Could you cast your eye over it, see what you think?

LindsayHill:
Related PR here https://github.com/StackStorm/ipfabric-docs/pull/159

I think this one needs to be merged before that one will pass CI

Kami:
LGTM, but haven't tried it locally yet.

Kami:
Same as in other repo - https://github.com/StackStorm/ipfabric-docs/pull/159#issuecomment-487097591

To be on the safe side, we should update cache key.

LindsayHill:
Going to pull in a punch of sphinx_rtd_theme updates into this before merging. We diverged from upstream a while back, and there's a lot of updates. Better to combine them with sphinx2, as there's mutual dependencies there.

LindsayHill:
Decided to get Sphinx + Py3 done as one change, will do further theme changes as separate PRs

PR #892, opened 2019-04-28
correct st2api restart command

punkrokk:


docs/source/upgrade_notes.rst
```diff
@@ -49,8 +49,7 @@ Upgrade Notes
       enable = True
       backend = enterprise
 
-  After you do that, you need to restart st2api service for changes to take affect- ``sudo st2ctl
-  restart-service st2api``.
+  After you do that, you need to restart st2api service for changes to take affect- ``sudo st2ctl restart st2api``.
```
Kami: Thanks - good catch.

It should actually be ``restart-component`` since restart will just restart all the services.

Kami: For the reference - https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L41

Kami:
Thanks again for catching and reporting this.

I updated the text and merged this into master.

PR #896, opened 2019-05-07
Remove U14, add U18

LindsayHill:
Ubuntu 14 support is removed with ST2 3.1.

This PR removes U14, adds U18 references, and breaks the existing U14/U16 install doc into two separate docs, as Mistral + Postgresql are not required on U18

General shape of docs here is correct, but will need some further testing to ensure all works as expected.


LindsayHill:
ewcocs builds will continue to fail until we merge https://github.com/StackStorm/ipfabric-docs/pull/165

docs/source/upgrade_notes.rst
```diff
@@ -677,302 +677,3 @@ Upgrade Notes
 
 * ``st2ctl reload`` now also registers rules by default. Prior to this release actions, aliases,
   sensors, triggers and configs were registered. Now rules are also registered by default.
-
-|st2| v1.6
```
armab: All good except of removing history behind the `st2` `v2.0`. That's still "history".

Any particular thinking behind cleaning that out or we can keep it for future?

LindsayHill: Part of the problem was I was clearing out the "all_in_one" doc. 

Rather than figuring out how to update all of the update notes, it seemed easier to just remove upgrade notes, since 1.6 is ~3 years old now. 

People can still go and access old versions of docs if they really need.

armab: Sounds reasonable :+1:

PR #902, opened 2019-05-15
Remove released feature from up coming list

jinpingh:
For task with items in Orquesta workflows has been released in st2 2.10.0.  Remove it from up coming list.


m4dcoder:
Can you also remove the line item for notify and mistral migration utility?

jinpingh:
> Can you also remove the line item for notify and mistral migration utility?
Are these features already in?
Removed.

PR #919, opened 2019-07-09
Fix the name of expected s3 object on deploy

m4dcoder:
The st2 and ewc docs websites are expecting the name of s3 objects (folder) to be the major.minor version number without the "v" prefix. The deploy step in circleci is modified to use sed to extract just the major.minor version from the circleci branch variable and use the result as the name of the s3 objects where the docs are uploaded to.


m4dcoder:
I will cherry pick this into v3.1 after CIs here passed and merged.

PR #921, opened 2019-07-19
Fix instructions for upgrading StackStorm from Community to Enterprise

armab:
Closes stackstorm/st2#4745

Use EWC install script which takes care of RBAC roles and can be applied on top of existing st2 community edition.

### DIFF
--- a/docs/source/install/ewc.rst
+++ b/docs/source/install/ewc.rst
@@ -32,32 +32,13 @@ Upgrading from Community
 ------------------------
 
 Already have a working StackStorm system, and want to add |ewc|? No problem! No need to install a
-new system. You can install |ewc| on top of your existing system. Just run these commands, again
+new system. You can install |ewc| on top of your existing system. Just run install commands, again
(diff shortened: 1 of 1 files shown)


docs/source/install/ewc.rst
```diff
@@ -32,32 +32,13 @@ Upgrading from Community
 ------------------------
 
 Already have a working StackStorm system, and want to add |ewc|? No problem! No need to install a
-new system. You can install |ewc| on top of your existing system. Just run these commands, again
+new system. You can install |ewc| on top of your existing system. Just run install commands, again
```
blag: You might want to mention that they have to use the password for their st2admin user if they have changed it.

armab: Surprisingly, password doesn't matter for EWC upgrade:
https://github.com/StackStorm/bwc-installer/blob/60c29f1fdc8abe223e4c7e81cd2751ed2e62b594/scripts/bwc-installer-deb.sh#L54
as it's not involved in any upgrade operations.

Keeping an old username is however, important.

PR #925, opened 2019-07-25
Update orquesta doc sections

m4dcoder:
Update roadmap for v3.2, v3.3, and backlog. Add orquesta development section to docs. Add doc section for runtime context and functions. Update orquesta authoring section with a better example.

### DIFF
--- a/docs/source/roadmap.rst
+++ b/docs/source/roadmap.rst
@@ -28,6 +28,7 @@ contributions. Here's our plans for the next two releases.
 3.3
 ---
 
+* **Ubuntu:** Stop producing Ubuntu 14.04 packages.
 * **Mistral deprecation:** Orquesta replaces Mistral as the workflow engine.
 
 Monitor the `master branch <https://github.com/StackStorm/st2/>`_ to see how we're progressing.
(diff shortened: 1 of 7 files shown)


docs/source/roadmap.rst
```diff
@@ -16,34 +16,30 @@ contributions. Here's our plans for the next two releases.
     the feature. Pull Requests are open to anyone.
 
 
-3.1
```
armab: As we released `3.1`, the implemented items from that release needs be moved under `Release History`

docs/source/roadmap.rst
```diff
@@ -16,34 +16,30 @@ contributions. Here's our plans for the next two releases.
...
-* **ChatOps:** RBAC.
-* **SAML:** Support SAML authentication.
+* **ChatOps:** Support RBAC.
+* **SSO:** Support SSO with SAML2 for |ewc| web UI (beta).
 * **Job Scheduling:** Job scheduling for ad-hoc jobs.
 
+3.3
+---
+
+* **Mistral deprecation:** Orquesta replaces Mistral as the workflow engine.
```
armab: :+1:

docs/source/roadmap.rst
```diff
@@ -16,34 +16,30 @@ contributions. Here's our plans for the next two releases.
     the feature. Pull Requests are open to anyone.
 
 
-3.1
----
-
-* **Ubuntu:** GA Support Ubuntu 18.04, with Python 3.6
-* **MongoDB:** Support MongoDB 4.0 (required for Ubuntu 18.04).
-* **Ubuntu:** Drop Ubuntu 14.04 support.
```
armab: Just so you're aware, we didn't really deprecate U14 support in `v3.1`, see https://github.com/StackStorm/st2/issues/4747. To avoid moving it under the "History" section.

m4dcoder: I'm going to communicate in the roadmap that we drop support of u14 in 3.1 but packages are still available.  I also added an entry in v3.3 to stop producing anymore u14 packages.

docs/source/roadmap.rst
m4dcoder: Done

PR #928, opened 2019-07-26
Add orquesta online evaluator as an external tool

trstruth:


docs/source/orquesta/start.rst
```diff
@@ -201,3 +201,4 @@ Additional Tools and Resources
   <http://yaqluator.com/>`_
 * `Jinja2 template engine documentation <http://jinja.pocoo.org>`_ and `Jinja2 online evaluator
   <http://jinja.quantprogramming.com/>`_
+* `Orquesta online evaluator <http://www.orquestaedit.com/>`_
```
m4dcoder: Maybe say `Workflow visualization tool <http://www.orquestaedit.com/>`_?   Also, since you own the tool, can you add a line or two instruction what user needs to do?  I go to the site and just serves me a blank page.  I assume just copy and paste the WF definition there.  But would be nice to have minor guidance.  Thanks

trstruth: Hey I changed the wording as you suggested, and updated the site to display some tool tips:
image

what do you think?

m4dcoder: Yeah.  better.  it's sufficient.

PR #934, opened 2019-09-20
Add section on pack dependencies

blag:
Documentation for https://github.com/StackStorm/st2/pull/4769, which adds support for specifying pack dependencies to `pack.yaml`.

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -133,6 +133,35 @@ version or **upgrade to latest** if the version is not specified. Your config fi
 overwritten, so you can revert to an older version just as easily, but for production deployments
 we recommend to always specify versions in case there are major changes in ``latest``.
 
+Pack Dependencies
+~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    New feature! As of StackStorm 3.2.
+
+If your pack uses actions from other packs, you can specify them in the ``dependencies`` section of
+the ``pack.yaml`` file and StackStorm will install them automatically when installing your pack.
+
+Similar to using the ``st2 pack install`` subcommand, you can reference packs from StackStorm
+Exchange using just their names, or you can specify a pack's Git repository URL. You can also use
+the same syntax to install a specific version, tag, or branch:
+
+.. code-block:: yaml
+
+    dependencies:
+      - excel
+      - powerpoint=0.2.2
+      - https://github.com/StackStorm/stackstorm-ms.git
+
+If you have dependency conflicts, the ``st2 pack install`` subcommand may error out, without
+installing any packs. If you would like to forcibly install a pack without installing its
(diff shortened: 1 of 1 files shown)


docs/source/packs.rst
```diff
@@ -133,6 +133,35 @@ version or **upgrade to latest** if the version is not specified. Your config fi
...
 we recommend to always specify versions in case there are major changes in ``latest``.
 
+Pack Dependencies
+~~~~~~~~~~~~~~~~~
+
+.. note::
+
+    New feature! As of StackStorm 3.2.
+
+If you pack uses actions from other packs, you can specify them in the ``dependencies`` section of
```
punkrokk: `If your pack`

blag: Fixed, thanks!

docs/source/packs.rst
```diff
@@ -133,6 +133,35 @@ version or **upgrade to latest** if the version is not specified. Your config fi
...
+
+If you have dependency conflicts, the ``st2 pack install`` subcommand may error out, without
+installing any packs. If you would like to forcibly install a pack without installing its
+dependencies, you can use the ``--skip-dependencies`` flag:
+
+.. code-block:: bash
+
+    st2 pack install --skip-dependencies my-custom-pack
+
 To uninstall a pack, use ``remove``:
```
armab: This is part of `Installing a pack` section.
Looks like the new `Pack Dependencies` section breaks it into 2 pieces.

Would be better to move `Pack Dependencies` after pack uninstall instructions.

m4dcoder: I think it makes sense to be part of or after pack install section.  Otherwise, folks will just miss this.

blag: Yeah, if anything, we should move the pack uninstallation instructions out of the *installation* section. This is like the old *Start* menu -> Shutdown option in Windows.

But that is outside the scope of this PR.

armab: With the way how `Pack Dependencies` was added, there is no separation and so Uninstall is part of Dependencies header/section.

I'm good if we place this under the new Uninstall section.

blag: Pack dependencies only relate to installing a pack though. If I move the subsection below the uninstallation instructions, I think people will miss the dependency section entirely.

armab: `Uninstall` can be somewhat related to `Install` section where removal of a pack comes as a conclusion. But it's not a part of `Pack Dependencies` which primary topic is special case about the dependencies syntax/functionality.

To fix this or we move that small pack removal note back to Install or add additional `Uninstall` sub-header. I'm good with any of these.

blag: Fixed in https://github.com/StackStorm/st2docs/pull/936.

PR #949, opened 2019-12-16
Update note on supported MongoDB versions

Kami:
I updated a section which was out of date.

The code base has supported v3.6 and 4.0 for a while now (and 4.0 is used as a default on Ubuntu Bionic).


punkrokk:
I’d like to add this the the CentOS 8 project, update this PR to include EL 8 and merge it when the EL 8 packages get released. Unless you want a new PR?

punkrokk:
Python 3 support for mongo is added in 4.2 -> Python 3.7. Boosts the Python 3.7 case.

armab:
@Kami  From what I remember the performance issues with MongoDB 3.6 are now fixed in upstream, do you remember where is the original issue to confirm?

armab:
@punkrokk Not sure about py3 vs MongoDB 4.2 and what do you mean exactly. How mongodb requires python as a dependency, can you provide more info?

MongoDB 4.0 is used with StackStorm running python 3.6 under Ubuntu Bionic installation. Also pymongo (st2 pip dependency) works OK in py 3.5+: https://api.mongodb.com/python/current/python3.html

punkrokk:
@armab from the mongo docs, while it appears everything works with mongodb 4.0 and CentOS 8, the docs say official EL8 support starts with Mongo 4.2: https://docs.mongodb.com/manual/tutorial/install-mongodb-on-red-hat/#considerations

docs/source/install/__mongodb_note.rst
```diff
@@ -1,5 +1,5 @@
 .. note::
 
-  The currently supported version of MongoDB is 3.4 and 4.0. This is the version installed by
+  The currently supported versions of MongoDB are 3.4 and 4.0. This is the version installed by
   the installer script. MongoDB 4.0 is installed by default on Ubuntu 18.04 and RHEL/CentOS8.
   MongoDB 3.6 does not currently work with StackStorm. Support for MongoDB 4.0 was added in ST2 v3.1.
```
armab, line 5: I think the original note with message about `mongodb 3.6` was a bit different:

```
  The currently supported version of MongoDB is 3.4. This is the version installed by the installer
  script.

  MongoDB 3.6 and 4.0 are also supported by StackStorm >= 3.0.0, but we have observed some
  performance regressions with MongoDB 3.6 so the default version which is installed on Ubuntu
  Xenial (16.04) still is 3.4.

  Right now MongoDB 4.0 is only used with Ubuntu Bionic (18.04).
```

punkrokk:
Which do you prefer @arma?

armab:
@punkrokk Check the https://github.com/StackStorm/st2docs/commit/4a65ac5ceeaa2cd7eea2928e5ce27bec53cb83f7 vs https://github.com/StackStorm/st2docs/commit/02d2a0901964c174e8b29899f1ac81e81ebd9f62

I think huge part of the text/diff from this PR was lost during the git conflict.

armab:
https://github.com/StackStorm/st2docs/pull/968 should fix it.

PR #954, opened 2020-01-12
New example for an Orquesta workflow querying multiple parameters

winem:
This PR depends on https://github.com/StackStorm/st2/pull/4841 which needs to be merged first.

This will show an extended version of the orquesta-ask-basic.yaml that queries multiple parameters and shows how to use the different values from the response.


punkrokk:
Looks like the example has been merged so this can be?

PR #960, opened 2020-02-05
document number of workers per service

punkrokk:
Add note to HA docs to document how to increase the number of workers in the Action Runner service.

Credit to @nmaludy: https://github.com/StackStorm/puppet-st2/blob/master/templates/etc/sysconfig/st2actionrunner.erb


docs/source/reference/ha.rst
```diff
@@ -162,6 +162,14 @@ In a proper distributed setup it is recommended to setup Zookeeper or Redis to p
 distributed co-ordination layer. See :doc:`Policies </reference/policies>`. Using the default
 file-based co-ordination backend will not work as it would in a single box deployment.
 
+By default the ``st2actionrunner`` service runs 10 workers. This number can be increased by
+modifying the environment variable ``WORKERS``. To make this permanent, create or edit
+``/etc/sysconfig/st2actionrunner`` and add the following line:
```
armab: It's not :100: accurate. The number of WORKERS are determined depending on number of CPU cores machine has.
See 
https://github.com/StackStorm/st2-packages/blob/1fb2f7a60e5d88895a5dfed79c55edc68482dc43/packages/st2/debian/st2.st2actionrunner.upstart#L30-L34

Additionally, `/etc/sysconfig/st2actionrunner` is for EL. For Ubuntu the respective config file to place custom ENV variables for the service/unit is `/etc/default/st2actionrunner`.

See https://docs.stackstorm.com/packs.html#proxy-configuration-via-environment-variables as example.

punkrokk: Updated accordingly

armab:
The text/description looks good :+1: 
I'm wondering if there is a better place for `st2actionrunner` configuration, outside of HA deployment? Setting number of workers is not just HA-specific, but valid for any st2 configuration.

@punkrokk @nmaludy  What do you think about including this info in https://docs.stackstorm.com/install/config/config.html page instead? And we can-cross link from HA doc to the main configuration doc.

docs/source/reference/ha.rst
armab: Yes, looks good :+1:

nmaludy:
@armab I thought the same thing but couldn't find the right place for it. I like your suggested location. We could then reference this from the HA docs?

punkrokk:
@nmaludy @armab I like the suggestion. I just pushed it. Just need to confirm the links work properly.

punkrokk:
any time!

PR #961, opened 2020-02-14
remove a t

punkrokk:
Found a rogue `t`. Squashed it.

PR #962, opened 2020-02-14
Remove --no-site-packages flag

blag:
It's now causing build failures, example: #961.

Incorporates and supercedes #961.


punkrokk:
Seems like —no-site-packages was deprecated as early as 2012. Can’t find the exact date.

PR #964, opened 2020-02-21
remove extra comma from json

punkrokk:
The current json for the example Graphite Dashboard has a comma after width in each section which is invalid JSON. I update the json to it works by removing the commas.


armab:
@punkrokk Looks like CI is not happy with these changes.

punkrokk:
Pushed EOF New line. Should pass in a moment

punkrokk:
pushed EOF New line, should pass shortly

punkrokk:
@armab Fixed, you'll see in the changes what I had to do.

PR #965, opened 2020-03-09
Update documentation on bastion ssh configs to use the proxycommand values

eedgar:


docs/source/troubleshooting/ssh.rst
```diff
@@ -85,7 +85,48 @@ runner boxes, and add the following configuration lines in ``/etc/st2/st2.conf``
...
+Make sure your ssh config is in the same account as user running the st2action process.  If root is running 
+st2actions install it there.  Make sure the config and identity files have proper permissions and ownership.
+ 
+.. code-block:: bash
+
+  chown -R root:root /root/.ssh/*
+  chmod 600 /root/.ssh/config
+  chmod 600 /root/.ssh/id_rsa
+
+This is a sample ssh config that is known to work with bastion forwarding.
```
punkrokk:
```suggestion
If you are looking to do ssh bastion forwarding, while allowing SSH to resolve automatically the correct keys based on hostname (eg. to dynamically support environments where a ssh hosts are set in your ssh_config file): 

This is a sample ssh config that is known to work with bastion forwarding.
```

punkrokk (changes requested):

docs/source/troubleshooting/ssh.rst
```diff
@@ -85,7 +85,48 @@ runner boxes, and add the following configuration lines in ``/etc/st2/st2.conf``
 
   [ssh_runner]
   use_ssh_config = True
-  ssh_config_file_path = /home/stanley/.ssh/config
+  ssh_config_file_path = /root/.ssh/config
+
+Make sure your ssh config is in the same account as user running the st2action process.  If root is running
```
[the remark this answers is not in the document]
eedgar: I was testing against the docker-stackstorm checkout... if prod is running the daemon than changing the owner for all files to stanley makes sense.  You may want to still include that the docker all in run runs as root or keep it running as the staction command owner.

arm4b (changes requested):
Initially this looks like a SSH troubleshooting doc.

Wouldn't it be better to include anything related to custom bastion SSH config in a separated sub-topic or even another page?

@punkrokk Do you think we have a better place for such configuration example?

punkrokk:
@armab I'm not opposed to this. Let's see what happens with the PR (https://github.com/StackStorm/st2/pull/4881) and then we can update and make it happen.

docs/source/troubleshooting/ssh.rst
```diff
@@ -85,7 +85,48 @@ runner boxes, and add the following configuration lines in ``/etc/st2/st2.conf``
...
+ 
+.. code-block:: bash
+
+  chown -R root:root /root/.ssh/*
+  chmod 600 /root/.ssh/config
+  chmod 600 /root/.ssh/id_rsa
+
+This is a sample ssh config that is known to work with bastion forwarding.
+
+.. code-block:: ini
```
blag, line 101: GitHub's [Linguist project](https://github.com/github/linguist) [supports syntax highlighting for SSH (client) configs](https://github.com/github/linguist/issues/3450), but you have to mark it as `ssh-config`:

```suggestion
.. code-block:: ssh-config
```

SSH (client) config coloring screenshot

And while you are at it, could you also fix this for line 160?

Leave line 84 alone though, since `/etc/st2/st2.conf` is just an INI file and is correctly marked as such. :+1:

blag (changes requested):

docs/source/troubleshooting/ssh.rst
cognifloyd: I just pushed a fix for this.

cognifloyd:
I just reworded some of the prose about bastion forwarding. Hopefully it fits better in the troubleshooting section now. If not, I'm not sure where else it might go.
@armab and @punkrokk What do you think, is this good to merge now?

cognifloyd (approved):

docs/source/troubleshooting/ssh.rst
cognifloyd: Ah. We can't use `ssh-config` because this Sphinx uses pygments, not GitHub's linguist. So `ini` is the way to go.

PR #968, opened 2020-04-14
Bring back original MongoDB 3.6 message from #949

armab:
Looks like https://github.com/StackStorm/st2docs/pull/949 came with a messed git conflict resolve resulting in a lost message about MongoDB v3.6 compatibility and performance.

See original diff https://github.com/StackStorm/st2docs/pull/949/commits/4a65ac5ceeaa2cd7eea2928e5ce27bec53cb83f7 comparing to #949

This PR fixes it.


punkrokk:
@armab thanks! Lo siento.

PR #969, opened 2020-04-14
Sync Roadmap for st2 v3.2

armab:
Sync roadmap for upcoming st2 `v3.2` and remove already finished items from the next `v3.3`.

@StackStorm/tsc please let me know if I'm missing something or anything else we can commit in future.

### DIFF
--- a/docs/source/roadmap.rst
+++ b/docs/source/roadmap.rst
@@ -16,28 +16,23 @@ contributions. Here's our plans for the next two releases.
     the feature. Pull Requests are open to anyone.
 
 
-3.2
----
-
-* **RHEL/CentOS:** Support RHEL 8.x (assuming it has been released!)
-* **RHEL/CentOS:** Drop support for RHEL/CentOS 6.x.
-* **ChatOps:** Support RBAC.
-* **Orquesta:** Support task retry in workflow definition.
-* **Orquesta:** Support rerun of workflow execution from task(s).
-
 3.3
 ---
 
-* **Ubuntu:** Stop producing Ubuntu 14.04 packages.
+* **RHEL/CentOS:** Drop support for RHEL/CentOS 6.x.
 * **Mistral deprecation:** Orquesta replaces Mistral as the workflow engine.
+
 
 Monitor the `master branch <https://github.com/StackStorm/st2/>`_ to see how we're progressing.
 
 Backlog
 -------
 
-Here's some more things on our list that we haven't scheduled yet:
+Here's some more nice things on our list that we haven't scheduled yet:
 
+* **Python:** Drop Python 2 support, rely on Python 3 ony.
+* **ChatOps:** Support RBAC.
+* **Docker:** Overhaul for `docker-st2 <https://github.com/stackstorm/st2-docker>`_.
 * **Workflow runtime graph:** Runtime view of workflow execution in st2flow for |ewc|.
 * **Workflow dry run:** Ability to run unit tests on orquesta workflows for |ewc|.
 * **SSO:** Support SSO with SAML2 for |ewc| web UI (beta).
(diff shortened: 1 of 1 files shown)


docs/source/roadmap.rst
```diff
@@ -65,6 +60,14 @@ Submit a PR!
 Release History
 ---------------
 
+.. rubric:: Done in v3.2
+
+* **RHEL/CentOS:** Support CentOS 8/RHEL 8 with Python 3.6 and MongoDB 4.0.
+* **Ubuntu:** Stop producing Ubuntu 14.04 packages.
```
nmaludy, line 66:
```suggestion
* **Ubuntu:** Drop support for Ubuntu 14.04.
```
Thoughts?

armab: We dropped support in `v3.1`, see `Done in v3.1` section.

docs/source/roadmap.rst
```diff
@@ -16,28 +16,23 @@ contributions. Here's our plans for the next two releases.
...
+
 Monitor the `master branch <https://github.com/StackStorm/st2/>`_ to see how we're progressing.
 
 Backlog
 -------
 
-Here's some more things on our list that we haven't scheduled yet:
+Here's some more nice things on our list that we haven't scheduled yet:
 
+* **Python:** Drop Python 2 support, rely on Python 3 ony.
```
armab, line 33: @blag RE `py2` vs `py3`.
There were some scattered messages, but I don't remember these kind of agreed plans for any release.
Sounds like it's a good topic for the next TSC meeting. Please raise it. I like the goal too :+1: 

Removing py 2.7 won't be that easy as it'll affect `U16` and `EL7` and packaging we currently use.
Probably worth a more in-depth technical @StackStorm/tsc conversation what's technically involved there and commitments around this item.

docs/source/roadmap.rst
blag: We also dropped all testing of the Ubuntu 14.04 packages. Ideally, every packaged environment is also tested and officially supported, but that's not always the case, and it wasn't the case here.

I don't know how relevant that will be to any/many users though.

docs/source/roadmap.rst
blag: Gotcha, that makes sense. I'll add it to the docket for the next TSC meeting.

docs/source/roadmap.rst
armab: To recap. It's possible to stop supporting the OS/distro, while still shipping the packages. This means no support, help, bugfixes provided as well as no issues accepted related to that distro/installation.

We added U18 and dropped U14 support in `v3.1`. By shipping the U14 packages one more release after `v3.1` we just gave some time to someone who was still relying on that but couldn't switch to other supported OS/distro.

And so as result Roadmap resulted in the following:
```
## Done in v3.2
**Ubuntu:** Stop producing Ubuntu 14.04 packages.

## Done in v3.1
**Ubuntu:** Drop Ubuntu 14.04 support. Packages are still available for a limited time.
**Ubuntu:** GA Support Ubuntu 18.04, with Python 3.6
```

@nmaludy I hope that makes more sense now.

nmaludy: Awesome!

nmaludy:
@armab FYI i just rebased this with master for you

docs/source/roadmap.rst
```diff
@@ -16,28 +16,24 @@ contributions. Here's our plans for the next two releases.
...
+* **ChatOps:** Support RBAC.
+* **Ubuntu:** Support Ubuntu 20.04 LTS.
```
armab, line 35: @blag Following your idea, just included Ubuntu 20.04 LTS to the list of items that's not assigned to any release yet.

blag: Awesome, thanks!

PR #973, opened 2020-04-21
Update pack documentation to state it uses last tagged commit

amanda11:
Minor update to the pack documentation to state that it uses the last 'tagged' commit rather than just last commit - new for v3.2

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -102,7 +102,7 @@ Exchange, and you can install your own packs from git just as easily.
     # Install your own pack from git
     st2 pack install https://github.com/emedvedev/chatops_tutorial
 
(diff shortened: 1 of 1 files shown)


docs/source/packs.rst
```diff
@@ -102,7 +102,7 @@ Exchange, and you can install your own packs from git just as easily.
     # Install your own pack from git
     st2 pack install https://github.com/emedvedev/chatops_tutorial
 
-By default, the latest commit to a pack repository will be installed, but you can specify a particular
+By default, the latest tagged commit to a pack repository will be installed, but you can specify a particular
```
armab, line 105: _Tagged commit_ looks a bit weird, how about this?

```suggestion
By default, the latest release from the pack repository will be installed, but you can specify a particular
```
Any better examples?

PR #982, opened 2020-05-16
Small doc typo - verb as syntax

relishcolouredhat:
code block included verb 'run' as if part of syntax


LindsayHill:
Thanks. Don't worry about the build failures, that's caused by something else going on, not related to your changes. Once that gets fixed this can get merged

PR #983, opened 2020-05-21
Update docker set-up -- seemed like  missing steps

timfong888:
I got stuck at the end of the docker install set up, so went to the README  and found more to get unblocked.

It seemed to just end at starting the container, but the install tutorials don't pick up on how to execute at the command line within the container.

### DIFF
--- a/docs/source/install/docker.rst
+++ b/docs/source/install/docker.rst
@@ -129,7 +129,7 @@ To stop the containers, run:
 
 To run from the command line inside the container:
 
-.. sourcecode::bash
+.. sourcecode:: bash
 
   docker-compose exec stackstorm bash
   
--- a/docs/source/install/docker.rst
+++ b/docs/source/install/docker.rst
@@ -135,6 +135,6 @@ To run from the command line inside the container:
   
 The installation tutorial can continue from here.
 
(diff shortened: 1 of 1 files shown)


docs/source/install/docker.rst
```diff
@@ -127,3 +127,14 @@ To stop the containers, run:
 
   docker-compose down
 
+To run from the command line inside the container:
+
+.. sourcecode::bash
```
armab, line 132: Looks like this block is not rendered well in rst: https://github.com/StackStorm/st2docs/blob/c6025ba2152bddcac747adb7b8a540c5bcb83797/docs/source/install/docker.rst

There seems to be formatting issue here:

```suggestion
.. sourcecode:: bash
```

docs/source/install/docker.rst
```diff
@@ -127,3 +127,14 @@ To stop the containers, run:
...
+
+  docker-compose exec stackstorm bash
+  
+The installation tutorial can continue from here.
+
+To run the UI, look for the password in `st2-docker/conf/stackstorm.env` and run http://localhost
```
armab, line 138:
```suggestion
To run the UI, look for the password in ``st2-docker/conf/stackstorm.env`` and run http://localhost/
```
rst needs a double `` to highlight that properly.

PR #986, opened 2020-06-06
Considering nginx sites-enabled

lat0z:
Added the consideration of the sites-enabled default file to avoid further errors

### DIFF
--- a/docs/source/install/u18.rst
+++ b/docs/source/install/u18.rst
@@ -131,7 +131,7 @@ SSL termination, and reverse-proxy st2auth and st2api API endpoints. To set it u
   # Remove default site, if present
   sudo rm /etc/nginx/conf.d/default.conf
   # Check for a default site on sites-enabled to avoid a duplicate default server error 
-  sudo rm /etc/nginx/sites-enabled/default
+  sudo rm -f /etc/nginx/sites-enabled/default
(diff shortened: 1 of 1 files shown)


punkrokk:
Is this true on the most recent version of nginx? E.g. what we use with CentOS 8?

lat0z:
On ubuntu 18.04 and 20.04 checking with nginx 1.14 and 1.17 there is the sites-enabled/default file, on centos 8 nginx 1.14 and 1.19 there is not

docs/source/install/u18.rst
```diff
@@ -130,6 +130,8 @@ SSL termination, and reverse-proxy st2auth and st2api API endpoints. To set it u
 
   # Remove default site, if present
   sudo rm /etc/nginx/conf.d/default.conf
+  # Check for a default site on sites-enabled to avoid a duplicate default server error 
+  sudo rm /etc/nginx/sites-enabled/default
```
LindsayHill, line 134: Could use `sudo rm -f`, so it doesn't complain if the file doesn't exist?

lat0z:
I fixed the rm command following your recommendation @LindsayHill

LindsayHill:
Thanks @lat0z

PR #989, opened 2020-06-26
Update supported ansible platforms

amanda11:
Updated st2docs to reflect that Centos8 support has been added, and that Ubuntu Trusty support is removed under PR https://github.com/StackStorm/ansible-st2/pull/265.

Can't merge until https://github.com/StackStorm/ansible-st2/pull/265 is completed.


docs/source/install/ansible.rst
```diff
@@ -18,10 +18,10 @@ Supported Platforms
 
 Our Ansible playbooks support the same platforms as manual installation, i.e.:
 
-* Ubuntu Xenial (16.04)
```
LindsayHill, line 21: I don't think you want to remove Xenial from here? Should just be Trusty removed (which it looks like we did a while back)

amanda11: Sorry about that - realised I'd missed updating docs when I had added CentOS 8. And then when removing Trusty in playbooks thought needed a doc update. But seems the doc update on ansible.rst is ahead of the actual playbooks.

Technically the playbooks don't I believe support Bionic yet. But I've not removed that unless you want it removed...

PR #999, opened 2020-07-27
Removing unnecessary references to mistral in U18 and RHEL8 install docs

gnewson:
Mistal isn't installed for RHEL8 or Ubuntu18 but it still appeared in the list of config files to edit under "Install ST2 Components".

As it's a shared section I created a new shared section without the references to Mistral and included that from both the RHEL8 and Ubuntu18 install docs.

### DIFF
--- a/docs/source/install/common/configure_components_no_mistral.rst
+++ b/docs/source/install/common/configure_components_no_mistral.rst
@@ -1,0 +1,7 @@
+If you are not running RabbitMQ, MongoDB on the same system, or have changed the
+defaults, please adjust these settings:
+
+* RabbitMQ connection at ``/etc/st2/st2.conf``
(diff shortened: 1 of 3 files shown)


amanda11:
In the 3.3 release then mistral is being removed for all operating systems. I'm wondering if its better to just put a note in for now on the existing shared components to state its not installed on U18 and EL8, as then there will be less documentation changes when we completely remove it.

LindsayHill:
Looks like part of the problem is related to https://github.com/StackStorm/st2/pull/4957, which probably needs to be back ported to the 3.2 branch to fix this build. 

This is fixed in master branch, but not here. Is a problem for building older versions of docs.

Could either be fixed in ST2 3.2 branch, or hopefully we can just get v3.3 published soon, and this goes away

PR #1000, opened 2020-07-28
Update Docker docs

armab:
Update Docker deployment documentation following the overhaul in https://github.com/stackstorm/st2-docker (https://github.com/StackStorm/st2-docker/pull/192).


armab:
Once merged, I'll also need to create the same PR against the `v3.2` stable branch to update https://docs.stackstorm.com/install/docker.html

PR #1011, opened 2020-08-14
Removal of st2mistral from basic chapter of docs

amanda11:
Removal of st2mistral from the Basic chapter.

Addresses part of the st2docs section of StackStorm/st2#4762


amanda11:
Couple of points:

1. Chatops and notify and orquesta - was no section for this before, but the parameter was mentioned in the orquesta pages but no notify example. I have taken the example from the  st2/st2tests/st2tests/fixtures/packs/orquesta_tests/actions/notify.yaml but couldn't get to this to trigger notify events
2. I suggest this is not merged until after the PRs on install, references and troubleshooting as they all reference the mistral. This PR could be included, as it has taken a copy of files in the other PRs.

winem:
Looks good considering @amanda11 s recent comment.

docs/source/mistral_yaql.rst
```diff
@@ -1,326 +0,0 @@
...
-            task1:
-                with-items:
-                    - name in <% $.names %>
-                    - ip in <% $.ips %>
-                action: examples.create-vm
-                input:
-                    name: <% $.name %>
-                    ip: <% $.ip %>
-
-The sections below contain additional YAQL examples of how to work with lists and dictionaries,
```
amanda11, line 137: There is a lot of useful general info here on YAQL. In the Orquesta pages the approach seems to have been to just point to the YAQL and Jinja projects themselves?

Are we happy to remove the pages and just refer to the YAQL/Jinja projects or want to put a page together for Orquesta/YAQL and Orquesta/Jinja?

docs/source/chatops/notifications.rst
```diff
@@ -178,54 +178,65 @@ task notify is shown below:
...
+    pack: orquesta_tests
+    runner_type: orquesta
+    entry_point: workflows/sequential.yaml
+    enabled: true
+    parameters:
+      who:
+        required: true
+        type: string
+        default: Stanley
+      notify:
```
amanda11, line 200: This is the example used in the unit-tests, but I couldn't get notify in Orquesta working in my setup - but I may be missing something. In the Orquesta pages we just mention there is a notify parameter that we can ignore for now...

docs/source/mistral_yaql.rst
winem: I would vote for a page for Orquesta/YAQL and Orquesta/Jinja. But this shouldn't be a blocker for this PR imho.

armab:
@amanda11 now as you have Github repo permissions, as PR author feel free to merge it when you see the fit.

docs/source/chatops/notifications.rst
amanda11: I got Orquesta notifications working but I also had to add on a notify section as well as the notify parameters, e.g. I also had:
```
notify:
  on-complete:
    routes:
      - slack
    message: "Succeeded"
```

so that it had the details of route etc. 

Do we think the documentation as is is ok, or 
1) should I add in the notify section into the example
2) Should I add in a preceding sentence along following lines:

`The method for specify the route and conditions for when tasks in an Orquesta workflow is the same as ActionChain. You have a notify section in the action meta when registering.   `      

and should I put in a notify section

amanda11: @LindsayHill @armab Having got  notifications working in Orquesta now, I'm wondering if more needs to go into this 
 section on the chatops page. Or alternatively I was mis-understanding something.

amanda11: > @LindsayHill @armab Having got notifications working in Orquesta now, I'm wondering if more needs to go into this
> section on the chatops page. Or alternatively I was mis-understanding something.

So as not to confuse things, I will merge this PR. And generate a new PR with enhancements that I think will make Orquesta notifications even clearer.

armab: :+1: step by step

PR #1012, opened 2020-08-14
Deprecate mistral in chapter advanced topics

winem:
This PR removes / rewords all occurences of mistral in the Advanced Topics section and fixes some headlines on the References and Guides/Exposed Metrics site which caused the headlines to appear in the navigation on the left. 

Currently it looks like this: 
Selection_170

With the fixed headlines it looks like this:
Selection_173


winem:
The warning that causes CircleCI to fail can be ignored for now. The file will be gone once we merged the other mistral-deprecation-MRs.

PR #1009, opened 2020-08-13
Remove mistral from installation section of documents

amanda11:
Mistral has been removed from StackStorm as part of 3.3 roadmap, update the install chapter of the documentation to remove mistral and postgreSQL references.
Not included: puppet installation changes

Addresses part of the st2docs section of https://github.com/StackStorm/st2/issues/4762

PR #1016, opened 2020-08-19
Update st2-deployment-big-picture

winem:
This PR replaces the st2-deployment-big-picture.png 

Please let me know if you want me to reference the picture via the weblink (as suggested here: https://github.com/StackStorm/st2/issues/4762#issuecomment-675059112) instead of having it in the repository itself as we do now. I think it's a bonus to have as less external references (=> dependencies) as possible. 

There was no text update needed as this was already covered by #1009


armab (approved):
Thanks @winem and @blag!

PR #1017, opened 2020-09-02
Fix possibly misspelt word.

anirudhbagri:


docs/source/actions.rst
```diff
@@ -331,7 +331,7 @@ Action Registration
 
 To register a new action:
 
-1. Place it into the content location.
+1. Place it into the correct location. (i.e under pack's action folder)
```
armab, line 334: Thanks for the contribution!

The `content location` is correct wording in this situation, eg it means pack content.

Maybe we could clarify it a bit more: 
```suggestion
1. Place it into the pack content location.
```

armab (changes requested):

armab:
Merging, Thanks @anirudhbagri!

armab (approved):

PR #1018, opened 2020-09-03
Updated docs with st2-deployment multi-node diagram with mistral removed

amanda11:
Updated docs with @blag 's updated st2-deployment-multi-node diagram that removed mistral


armab (approved):
Thanks!

PR #1021, opened 2020-09-15
Update information on python version used by packs

### LINKED ISSUE #1019: Description of version of python used for packs implies it is always python 2
The section "Using Python 3 for Pack Python Virtual Environment" in https://docs.stackstorm.com/latest/packs.html looks like it is out of date.
It states that packs will use python 2 unless the --use-python3 variable is used, and for example states
"By default, the same Python 2.7.x binary which is used by all the StackStorm components and services is used for pack virtual environments."

As I understand it the actions will run on the 2.7 for Ubuntu 16 and CentOS 7, and 3 for CentOS 8 and Ubuntu 18. So it might be worth re-wording to indicate that they will use the same version of python as the ST2 defaults to.

It then ends with:
"If you want all the StackStorm components and services to also run under Python 3 you should look at the Ubuntu Bionic packages where everything runs under Python 3.6. These are in the “unstable” repositories for 3.0, and will be GA in 3.1. On Ubuntu Bionic only packs which work under Python 3 are supported."
which is a bit out of date, as we have released ST 3.2.

amanda11:
Update information on what python version is used by StackStorm, and add python 2 warning.

Resolves #1019

### DIFF
--- a/docs/source/packs.rst
+++ b/docs/source/packs.rst
@@ -283,7 +283,7 @@ default private key.
 
 Deploy keys are more secure than personal access tokens and can be configured on a per-repo basis.
 
-Using Python 3 for Pack Python Virtual Environment
+Python versions in Pack Python Virtual Environment
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. note::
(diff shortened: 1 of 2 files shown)


armab (approved):
:+1:

PR #1024, opened 2020-10-06
Update upgrade_notes to mention fail fast on workflows

amanda11:
Included a comment on upgrade notes about fail-fast on orquesta workflows.

Designed to explain the confusion on when workflows stop, raised in StackStorm/st2#5031.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -17,7 +17,11 @@ Upgrade Notes
     trigger:
       type: "core.st2.action.file_written"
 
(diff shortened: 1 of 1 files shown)


emptywee:
Yes, sounds good. Would be better if we had a choice in how we want the engine to behave (like I proposed to have `fail` and `terminate` and one would keep other branches running, while the other would terminate the workflow for good), but for now it is more clear to understand why your ex-mistral workflows converted to orquesta don't work the same way.

m4dcoder (changes requested):
I don't know why this specific feature deserves a mention in the upgrade notes. There are other Mistral -> Orquesta nuance that users should be aware as well. Why not just ask users to review the orquesta documentation in general?

m4dcoder (approved):

Issue #1038, opened 2020-11-27
RBAC doc update after feature transition enterprise -> st2 OSS core


armab:
`RBAC` functionality was integrated into the st2 core: https://github.com/StackStorm/st2/pull/5086
Update RBAC documentation to remove any mentions of EWC/Enterprise and sync it up.

* https://docs.stackstorm.com/rbac.html
* https://github.com/StackStorm/st2docs/blob/master/docs/source/rbac.rst

Additionally, new instructions should indicate how to enable the `rbac` via `st2.conf`:
https://github.com/extremenetworks/ewc-installer/blob/4ac4c34073998c5ffd8e6d746b7e0cb9b3b5d346/scripts/bwc-installer-deb.sh#L201-L204

As well as show the recommended default RBAC definitions:
https://github.com/extremenetworks/ewc-installer/blob/4ac4c34073998c5ffd8e6d746b7e0cb9b3b5d346/scripts/bwc-installer-deb.sh#L206-L222

cc @hnanchahal

PR #1039, opened 2020-11-30
Updating rbac documentation

hnanchahal:
Closes https://github.com/StackStorm/st2docs/issues/1038

### DIFF
--- a/docs/source/rbac.rst
+++ b/docs/source/rbac.rst
@@ -1,6 +1,12 @@
 Role Based Access Control
 =========================
 
+.. note::
+
+   Prior to StackStorm 3.3, Extreme Networks provided a commercial version of the StackStorm automation
+   platform which included Role Based Access Control (RBAC). As these enterprise features were donated to
+   the Linux Foundation, RBAC is now available in StackStorm Open Source since 3.4.
+
 Role Based Access Control (RBAC) allows system administrators to restrict users' access and limit
 the operations they can perform. For instance, you could give your database operator access only
 to the database-related actions.
@@ -223,6 +229,9 @@ There are some exceptions, described below:
 
 Enabling RBAC
 -------------
+
+To configure RBAC you will need to manually enable it in ``st2.conf`` and assign ``admin`` privileges to default 
+user ``stanley``.
 
 To enable rbac, add this section to ``/etc/st2/st2.conf``:
 
--- a/docs/source/rbac.rst
+++ b/docs/source/rbac.rst
@@ -1,14 +1,6 @@
 Role Based Access Control
 =========================
 
-.. note::
-
-  Prior to StackStorm 3.3, Extreme Networks provided a commercial version of the StackStorm automation platform which included Role Based Access Control (RBAC).
-
-Extreme Workflow Composer was donated to Linux Foundation, but those components are not yet integrated into the StackStorm release. Please visit https://stackstorm.com/2020/05/27/extreme-networks-donates-ewc-to-linux-foundation/ for more details.
-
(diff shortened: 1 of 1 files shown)


amanda11 (commented):
Changes look good. One minor comment.

Previously we mentioned that there was a need to add admin privileges to Stanley user. If we install the community RBAC does that get added by the bash installer or do you have to do that manually?

I think we should mention which happens.

docs/source/rbac.rst
```diff
@@ -232,18 +224,13 @@ There are some exceptions, described below:
 Enabling RBAC
 -------------
 
-If you installed |ewc| using the :doc:`one-line install </install/ewc>`, RBAC will be automatically
-enabled. It will assign the ``admin`` role to ``stanley`` and ``st2admin``. 
-
-If you installed |ewc| separately, by installing the ``bwc-enterprise`` package on top of |st2|, you
-will need to manually enable RBAC, and assign ``admin`` privileges to ``stanley``. It is not
-enabled by default. To enable it, add this section to ``/etc/st2/st2.conf``:
+To enable rbac, add this section to ``/etc/st2/st2.conf``:
```
armab, line 227: Yeah, let's be a little bit more explicit that these configuration steps should be done manually. I think that should cover the comment by @amanda11.

```suggestion
To make RBAC work you will need to manually enable it and assign ``admin`` privileges to default user ``stanley``.

To enable RBAC, add this section to ``/etc/st2/st2.conf``:
```

docs/source/rbac.rst
```diff
@@ -1,14 +1,6 @@
 Role Based Access Control
 =========================
 
-.. note::
-
-  Prior to StackStorm 3.3, Extreme Networks provided a commercial version of the StackStorm automation platform which included Role Based Access Control (RBAC).
```
armab, line 6: Let's keep this note as it looks useful and also mention that RBAC is now available in StackStorm Open Source since version `3.4`.

Maybe something along the lines?
> Prior to StackStorm ``3.3``, Extreme Networks provided a commercial version of the StackStorm automation platform which included Role Based Access Control (RBAC). As these enterprise features were donated to the Linux Foundation, RBAC is now available in StackStorm Open Source since ``3.4``.

cc @m4dcoder WDYT?

docs/source/rbac.rst
```diff
@@ -1,14 +1,6 @@
...
-
-Extreme Workflow Composer was donated to Linux Foundation, but those components are not yet integrated into the StackStorm release. Please visit https://stackstorm.com/2020/05/27/extreme-networks-donates-ewc-to-linux-foundation/ for more details.
-
-  These components are NOT available in the StackStorm 3.3 release, but will be integrated into the community edition of StackStorm in future release(s).
-
 Role Based Access Control (RBAC) allows system administrators to restrict users' access and limit
 the operations they can perform. For instance, you could give your database operator access only
 to the database-related actions.
```
armab, line 6: Same as in LDAP PR, I think it worth referring to https://github.com/stackstorm/st2-rbac-backend repository URL somewhere at start of the RBAC doc section so users will be able to find it and look at the source code if they need.

armab: @hnanchahal we'll still need to include the original RBAC repository URL in st2docs, see above.

That's only blocker for this PR.

amanda11:
I'm not sure what has happened, but this PR now only contains LDAP changes, rather than the RBAC ones @armab

armab:
Thanks @amanda11, fixed now!
I fat-fingered it today while trying to do several things in parallel. Rarely works well :)

armab (approved):
:+1:

PR #1041, opened 2020-12-04
Remove obsolete submit debug info tool

amanda11:
Remove obsolete submit-debug-info


armab (commented):
Good catch!

Though `st2-submit-debug-info` is really obsolete and not used for many years.
We can just remove all the documentation associated with it, as well as clean it up from the st2 codebase.

This is somewhat similar to what we did in another PR: [ remove obsolete check-licence #5092 ](https://github.com/StackStorm/st2/pull/5092)

amanda11:
> Good catch!
> 
> Though `st2-submit-debug-info` is really obsolete and not used for many years.
> We can just remove all the documentation associated with it, as well as clean it up from the st2 codebase.
> 
> This is somewhat similar to what we did in another PR: [ remove obsolete check-licence #5092 ](https://github.com/StackStorm/st2/pull/5092)

OK - I'll do the changes to remove from ST2 code base as well as docs.

armab (approved):
:+1:

PR #1043, opened 2020-12-11
Refactor docs building

blag:
I initially intended to just add a quick reference table for runner parameters (multiple people have requested this in #community), but in the process of building I ran into additional issues that I solved.

* [x] Add `orquesta` to `.dockerignore`
* [x] Update `Dockerfile` to use a pinned version of Ubuntu, 18.04, that has Python 3.6 (too new for Ubuntu 16.04 and too old for Ubuntu 20.04)
* [x] Improve instructions on how to regenerate documentation for runners
* [x] Refactor `Makefile` to remove enterprise make targets
* [x] Fix misapplied `PHONY` targets (should be `.PHONY`)
* [x] Refactored the `Makefile` to use real targets instead of simply running tasks - this greatly speeds up documentation improvements because you don't have to continuously install st2 and st2docs requirements
* [x] Fix a Python syntax
* [x] Add a quick reference table for parameters of common runners
* [x] Tweak the CSS to remove the `max-width` property so the table can display on the entire page


blag:
I don't think the `dir` parameter is used in `remote-shell-cmd` - we should remove it from `contrib/runners/remote_runner/remote_runner/runner.yaml`. It will then be removed from st2docs when we rebuild this documentation.

nmaludy (approved):
This looks awesome!

PR #1045, opened 2020-12-23
Remove st2resultstracker from the ha and monitoring reference docs

winem:
This PR completes the st2resultstracker deprecation and removes it from the docs.


armab (approved):
Looks good!

Thanks :+1:

PR #1046, opened 2021-01-05
Build from source updates

amanda11:
Update instructions to build from sources to include instructions for CentOS/RHEL and remove Fedora.

Updated Ubuntu instructions as they did not include python3.6-dev or the LDAP libraries.

### DIFF
--- a/docs/source/development/sources.rst
+++ b/docs/source/development/sources.rst
@@ -38,6 +38,38 @@ Fedora
     systemctl enable rabbitmq-server
     systemctl restart rabbitmq-server
 
+CentOS
+------
+
+.. code-block:: bash
+
+    yum install python-pip python-virtualenv python-tox gcc-c++ git-all screen icu libicu libicu-devel openssl-devel openldap-devel python3-devel
+
+    yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+
+    # Add key and repo for the latest stable MongoDB (4.0)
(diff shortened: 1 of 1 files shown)


docs/source/development/sources.rst
```diff
@@ -38,6 +38,34 @@ Fedora
     systemctl enable rabbitmq-server
     systemctl restart rabbitmq-server
 
+CentOS
+-----------
```
armab, line 42:
```suggestion
CentOS/RHEL
-----------
```

:+1: 

We also can remove the Fedora section above, as that was really meant yum-like systems.

docs/source/development/sources.rst
```diff
@@ -19,7 +19,7 @@ Ubuntu
 
 .. code-block:: bash
 
-    apt-get install python-pip python-virtualenv python-dev gcc git make realpath screen libffi-dev libssl-dev
+    apt-get install python-pip python-virtualenv python-dev gcc git make realpath screen libffi-dev libssl-dev python3-dev libldap2-dev libsasl2-dev
```
armab, line 22: I think `python-dev` is not needed anymore as that's replaced with the `python3-dev`

amanda11: Thanks for comments.
Tested on U16 as well, so have removed python-dev, and instead used python3.6-dev so works on both U16 and U18.

armab (approved):
:+1:

PR #1047, opened 2021-01-13
Include installation and v3.4 migration changes for Ubuntu Xenial + python3.6

armab:
Closes https://github.com/StackStorm/st2-packages/issues/672

### DIFF
--- a/docs/source/install/upgrades.rst
+++ b/docs/source/install/upgrades.rst
@@ -170,10 +170,10 @@ any skipped versions:
 v3.4
 ''''
 
-*  |st2| now uses python 3 on RHEL/CentOS 7. Therefore any packs that only support python 2 will need to be upgraded to python 3.
+*  |st2| now uses python 3 on Ubuntu 16 and RHEL/CentOS 7. Therefore any packs that only support python 2 will need to be upgraded to python 3.
 
 
-* RHEL 7.x only. Ensure python3-devel can be installed from an enabled repository:
+* *RHEL 7.x only.* Ensure python3-devel can be installed from an enabled repository:
 
   .. note::
 
@@ -198,6 +198,23 @@ v3.4
 
     sudo yum install python3-devel --enablerepo <optional-server-rpm repo>
 
+* *Ubuntu 16.04 Xenial only.* Python 3.6 is not available in the base Ubuntu Xenial distribution and you can add unofficial 3rd party `Python PPA repository <https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa>`_:
+
+  .. warning::
+
+     Please be aware of the support and security risks associated with using unofficial 3rd party PPA repository, StackStorm does NOT provide ANY support or security update for python3.6 packages on Ubuntu 16.04.
+     If security is a priority for you, we recommend starting migrating to Ubuntu 18.04 LTS (Bionic) as a base OS which has official python 3.6 packages.
+     This is a workaround to support Ubuntu Xenial with python 3 until we deprecate it in the future versions.
+
+  .. sourcecode:: bash
+
+    sudo apt-get install -y software-properties-common
+    # add 3rd party insecure python3 PPA repository
+    sudo add-apt-repository -y ppa:deadsnakes/ppa
+    sudo apt-get update
+
+    # ensure python3 package exists and could be installed
+    apt-cache show python3.6
 
 v3.3
 ''''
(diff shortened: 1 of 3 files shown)


blag (approved):

docs/source/install/upgrades.rst
```diff
@@ -198,6 +198,24 @@ v3.4
...
     sudo yum install python3-devel --enablerepo <optional-server-rpm repo>
 
+* *Ubuntu 16.04 Xenial only.* Python 3.6 is not available in the base Ubuntu Xenial distribution and you can add unofficial 3rd party `Python PPA repository <https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa>`_:
+
+  .. warning::
+
+     Please be aware of the support and security risks associated with using unofficial 3rd party PPA repository.
+     StackStorm does NOT provide ANY support or security update for python3.6 packages on Ubuntu 16.04.
+     If security is a priority for you, we recommend starting migrating to Ubuntu 18.04 LTS (Bionic) as a base OS which has official python 3.6 packages.
+     This is a workaround to support Ubuntu Xenial with python 3 until we deprecate it in the future versions.
```
amanda11, line 208: A question really / but I wondered why the apt-get show and the extra comment about this being workaround before xenial is dropped were only on the upgrade and not install page.
Just checking this was on purpose, rather than an oversight.

amanda11 (changes requested):
For the el7 changes we also added a comment to the upgrade-notes to state they need to migrate packs. We should probably add U16 to that sentence in the upgrade notes for 3.4

docs/source/install/upgrades.rst
```diff
@@ -170,10 +170,10 @@ any skipped versions:
 v3.4
 ''''
 
-*  |st2| now uses python 3 on RHEL/CentOS 7. Therefore any packs that only support python 2 will need to be upgraded to python 3.
+*  |st2| now uses python 3 on Ubuntu 16 and RHEL/CentOS 7. Therefore any packs that only support python 2 will need to be upgraded to python 3.
```
armab, line 173: @amanda11 Do you refer to this comment about packs migration or any other place?
I believe it applies for both U16 and EL7 now.

docs/source/install/upgrades.rst
armab: Right and good catch!

I thought that upgrade instructions should include a bit more info for the users to verify if the migration went well, while for raw installation instructions we just care about the end result without any debugging or transition steps or info.
That was an idea behind it.

docs/source/install/upgrades.rst
amanda11: Upgrade-notes.rat file, was what I was referring to.

docs/source/install/upgrades.rst
amanda11: That's good - makes sense, just wanted to double check

docs/source/install/upgrades.rst
armab: Thanks, I missed that!

Added in https://github.com/StackStorm/st2docs/pull/1047/commits/8ac1f0af4db8e4d4b141fb77e384bd856c756617, reworded it a little bit to accommodate the python 2 removal.

docs/source/upgrade_notes.rst
```diff
@@ -8,7 +8,9 @@ Upgrade Notes
 |st2| v3.4
 -------------
 
-* RHEL/CentOS 7.x ST2 distributions now uses python version 3 for ST2 and all packs, therefore any packs that only support python 2 will need to be migrated to python 3.
+* Python 2 support was removed.
+  Any packs that only support python 2 will need to be migrated to python 3.
+  Ubuntu Bionic 16.04 LTS and RHEL/CentOS 7.x ST2 distributions now uses python version 3.
```
amanda11, line 13: Really minor comment - but probably should be "use" instead of "uses"

amanda11 (approved):
Looks good - I left a really minor grammar comment on the upgrade-notes page.

docs/source/upgrade_notes.rst
armab: Cool, https://github.com/StackStorm/st2docs/pull/1047/commits/5ce861076b261a61b0d03142ae8e47d2286ce09a should fix it.

amanda11: Looks great, thanks

PR #1049, opened 2021-01-18
Correct interval for docs example

sigsevneo:
The code in `sensor1.py` is set to trigger every 60 seconds - https://github.com/StackStorm/st2/blob/71e0cdf4169a43556f12472811ff345d19217a06/contrib/hello_st2/sensors/sensor1.py#L22


armab (approved):
Thanks for fixing this oversight :+1:

PR #1057, opened 2021-02-28
Update the quick install command to support interactive scripts

winem:
The previous command had issues with interactive scripts. See https://github.com/StackStorm/discussions/issues/66#issuecomment-787154902 for an example (st2.conf was updated and the user was asked whether to keep/override the file). 

All options Y/I, N/O, Z and D were tested on Ubuntu 16.04 and 18.04.

### DIFF
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -53,7 +53,7 @@ on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run this comman
 
 .. code-block:: bash
 
-   curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password='Ch@ngeMe'
+   curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password=Ch@ngeMe
 
 This is an opinionated installation of |st2|. It will download and install all components, as per
 the :doc:`single host reference deployment <./overview>`. It assumes that you have a clean, basic
--- a/docs/source/install/index.rst
+++ b/docs/source/install/index.rst
@@ -53,7 +53,7 @@ on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run this comman
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/source/install/index.rst
```diff
@@ -53,7 +53,7 @@ on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run this comman
 
 .. code-block:: bash
 
-   curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password='Ch@ngeMe'
+   curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password=Ch@ngeMe
```
armab, line 56: Not sure how enclosing the password value with single quotes helps, considering it doesn't contain any special characters that might be interpreted in bash.

Any chance there were any other uncommitted changes?

docs/source/install/index.rst
```diff
@@ -53,7 +53,7 @@ on Ubuntu, or ``sudo yum install curl nss`` on RHEL/CentOS. Then run this comman
 
 .. code-block:: bash
 
-   curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password=Ch@ngeMe
+   bash <(curl -sSL https://stackstorm.com/packages/install.sh) --user=st2admin --password=Ch@ngeMe --unstable --u16-add-insecure-py3-ppa
```
blag: Are we sure we want to include `--unstable` here? Or is tweaking the `bash <(curl ...)` and adding the `--u16-add-insecure-py3-ppa` flag enough?

I'd like this change to be as small as humanly possible, since we are soooooo close to release.

winem: I actually removed the `--add-insecure-py3-ppa` and `--unstable` flag with the most recent commit. So we don't change anything on the behaviour or results but support interactive scripts now. I'm sorry for the confusion.

docs/source/install/index.rst
winem: This was a mistake and a commit was missing. I confused the configured remotes on my workstation.

blag: No worries, I've done the same. 😄

blag (approved):
LGTM

PR #1058, opened 2021-03-01
Update reload virtualenv command on upgrade

amanda11:
Update upgrade instructions following https://github.com/StackStorm/st2/pull/5167


cognifloyd (approved):
LGTM
Much nicer 😃

armab (approved):
Thanks for the follow-up and updating documentation once again!

PR #1063, opened 2021-04-15
Rabbitmq repo

amanda11:
Install latest version of rabbitmq from packagecloud, and therefore as that requires newer version of erlang, install the minimal erlang provided by rabbitmq - as per instructions at https://github.com/rabbitmq/erlang-rpm

Update install instructions to install erlang from rabbitmq_erlang. Also update the uninstall to include removal of the rabbitmq and erlang repos.

Relates to StackStorm/st2-packages#699


Kami:
Should we also document (if not ready) minimum supported and recommended RabbitMQ version somewhere in the docs?

Kami (approved):

PR #1074, opened 2021-06-10
Xenial removal and roadmap update

amanda11:


docs/source/roadmap.rst
```diff
@@ -64,6 +58,13 @@ Submit a PR!
 Release History
 ---------------
 
+.. rubric:: Done in v3.5
+
+* **Ubuntu Focal:** Add support for Ubuntu 20.04, with Python 3.8
```
winem, line 63: Do we want to mention mongo 4.4 here?

winem:
https://github.com/StackStorm/st2docs/blob/master/docs/source/install/ansible.rst
https://github.com/StackStorm/st2docs/blob/master/docs/source/install/puppet.rst

Both of them mention U16 and the puppet link even U14. 

Do we want to remove this in this PR or just when the respective PRs in the ansible-st2 and puppet repository got merged? Beside that it else looks good to me!

winem (approved):

amanda11:
I was going to go for updating puppet and ansible in docs - when we update the relevant installers.

docs/source/roadmap.rst
amanda11: Good point - updated

PR #1075, opened 2021-06-10
Add coordination service requirement for single node and HA install

m4dcoder:
The coordination service is required to support multiple workflow use cases. Update docs to let users know that redis is installed by default in single node installation script and that redis/zookeeper needs to be included in a HA deployment.

### DIFF
--- a/docs/source/coordination.rst
+++ b/docs/source/coordination.rst
@@ -1,0 +1,51 @@
+Coordination
+============
+
+Coordination backend is required to support workflows with multiple branches or tasks with items and
+actions with concurrency policies defined.
+
+StackStorm utilizes the ``OpenStack Tooz`` library for communicating with the coordination backend.
+The coordination backend must support the ``Locking`` functionality as defined by the ``Tooz`` 
+interface. Please refrence the `OpenStack Tooz compatability page <https://docs.openstack.org/tooz/latest/user/compatibility.html>`_
+for more information what interfaces are implemented by various backends.
+
+The following is a list of backends that can be configured for the coordination service. For the
+full list of the supported backends and how to configure them, please visit
+`OpenStack Tooz documentation <https://docs.openstack.org/tooz/latest/>`_.
+ 
+ * Redis
+ * Zookeeper
+ * consul
+ * etcd
+ * file (for testing when all the services are running on a single host)
+
+The configuration of the coordination service is done in the ``coordination`` section
+of ``/etc/st2/st2.conf``. The following are configuration examples for Redis and Zookeeper.
+
+Redis:
+
+.. code-block:: ini
+
+    [coordination]
+    url = redis://:password@host:port
+
+ZooKeeper:
+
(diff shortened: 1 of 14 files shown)


amanda11 (changes requested):
I think there are a few other places that need updating for Redis:

- Manual installation instructions - need to state to install redis - so u18, u20, el7 and el8 files
- Upgrade_notes - mention redis added
- Upgrades - in the version specific upgrade, need to supply instrutions to install redis prior to upgrade of st2 package
- Uninstall - uninstall redis steps as per the other steps to uninstall the other dependencies
- install/system_Requirements - redis port probably should be added to the list of ports used by default

docs/source/install/upgrades.rst
```diff
@@ -193,6 +193,12 @@ v3.5
      sudo rpm -e --nodeps nodejs
      sudo yum upgrade st2chatops
 
+* Redis server is installed and configured as backend for the coordination service
+  by default in the single node installation script to support workflows with multiple
+  branches and tasks with items. Upgrade requires coordination service to be setup
+  manually, For workflows to be executed properly, setup the coordination service
+  accordingly. See :doc:`../reference/policies` for setup instruction.
```
amanda11, line 200: The policies page doesn't really give instructions on how to install redis, just how to configure ST2 to point to it. So I don't know if we need to just put in here that you need to install a coordinations service, and then go to policies to configure it.

and/or maybe tell them that the manual installation instructions include steps to install redis which on a fresh install is now installed by default.

amanda11 (approved):
LGTM - just a comment on the upgrade that we might need to make it clearer about installing coordination as the linked page on policies hasn't got instructions for installing coordination service (but we do have that in the manual) - so I'm not sure if we need to give a little more info e.g. either give them the redis installation examples (cut/paste from the install) or just point them as well to refer to the redis commands in the manual install instructions.

But I'll leave that to your discretion.

docs/source/install/upgrades.rst
m4dcoder: I separated it out to a coordination section in below commit.

docs/source/coordination.rst
```diff
@@ -0,0 +1,51 @@
...
+Some of these coordination backends also require corresponding client libraries to be installed
+in |st2| virtualenv. We do not ship these libraries by default. As an example, to install the client
+library in |st2| virtualenv, run:
+
+.. sourcecode:: bash
+
+    sudo su
+
+    # Example when using redis backend
+    /opt/stackstorm/st2/bin/pip install redis
```
armab, line 48: Redis pip dependency is already included in the st2 requirements as a default coordination backend and in st2.conf.
https://github.com/StackStorm/st2/blob/cf19864cb7a89bc5131caaf4b0013311640e042a/st2common/requirements.txt#L34

It worth mentioning that, otherwise user assuming he needs to configure it further.

PR #1081, opened 2021-08-02
Add a warning about duplicate keys in rule criteria

blag:
PyYAML has some unfortunate behavior when it comes to duplicate key values, so specifying the same criteria key multiple times in a rule ignores all but the last criterion:

```yaml
criteria:
  trigger.payload.commit.tags:  # non-last duplicate key - ignored
    type: contains
    pattern: StackStorm
  trigger.payload.commit.tags:  # last duplicate key - evaluated
    type: ncontains
    pattern: pull request
```

This PR notes the workaround is to use an alternative syntax to specify trigger data:

```yaml
criteria:
  trigger.payload.commit.tags:
    type: contains
    pattern: StackStorm
  trigger.payload.commit['tags']:
    type: ncontains
    pattern: pull request
```

Hopefully this will help clear up some confusion for users.

### DIFF
--- a/docs/source/rules.rst
+++ b/docs/source/rules.rst
@@ -153,6 +153,50 @@ The ``pattern`` value can also reference a datastore value using Jinja variable 
 In this example we are referencing the value of a datastore item with the name
 ``current_build_number``.
 
+.. warning::
+
+    Each criteria key must be unique.
+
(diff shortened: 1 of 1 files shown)


amanda11 (approved):
LGTM

PR #1084, opened 2021-08-05
Add global permissions to documentation

### LINKED ISSUE #1083: Global RBAC permissions needs to be documented
Carrying part of the conversion over from https://github.com/StackStorm/st2/issues/5303#issuecomment-892134776

The global RBAC permission types are not documented. While there is some overlap for types within some of the resources there should be a section specifying the global settings as well.

amanda11:
Extract and include global permissions to rbac documentation.

Resolves #1083


scripts/generate-available-permission-types-table.py
```diff
@@ -46,31 +47,37 @@
     ResourceType.TRACE,
 ]
 
+def add_lines_for_permission_set(resource_title, permission_types, lines):
+    lines.append('%s' % (resource_title))
```
cognifloyd, line 51: Why format a single var like this?
```suggestion
    lines.append(resource_title)
```
Or if we do need to format:
```suggestion
    lines.append(f"{resource_title}")
```

scripts/generate-available-permission-types-table.py
```diff
@@ -46,31 +47,37 @@
...
+    lines.append('~' * len(resource_title))
+    lines.append('')
+
+    rows = []
+    rows.append(TABLE_HEADER)
+
+    for permission_type in permission_types:
+        description = PermissionType.get_permission_description(permission_type)
+        rows.append(['**%s**' % (permission_type), description])
```
cognifloyd, line 60: I prefer reading f strings. 😉 
```suggestion
        rows.append([f"**{permission_type}**", description])
```

scripts/generate-available-permission-types-table.py
cognifloyd: Oh. I see, you just moved what was already there.

cognifloyd (approved):
A couple of suggestions on possible style improvements.

Also, it looks like the python hasn't been formatted by black. Odd. We should fix that at some point. 🤷

blag:
This is probably a great time to update the string formatting that is touched by this PR to use f-strings, but otherwise looks good.

amanda11:
> This is probably a great time to update the string formatting that is touched by this PR to use f-strings, but otherwise looks good.

I got really confused - as I had done a change to put in f-strings as per @cognifloyd comment. However I hadn't pushed it back to git!!!

armab (approved):
Nice! :tada:

PR #1085, opened 2021-08-10
Added doc for modification of action delete API

mahesh-orch:
Adding documentation for modification of action delete API to remove related files on disk and added arguments `-f`, `--force` as auto delete flags to action delete command on CLI to remove disk files without asking user for permission.


blag:
@mahesh-orch Would you please use more descriptive commit messages, and minimize the number of commits you generate? For every single commit that you push, many people get an email from GitHub including the commit message, and a bunch of commits becomes a deluge of email notifications.

mahesh-orch:
@blag Sure, I will keep commits minimum and descriptive. Thanks.

amanda11:
@mahesh-orch For the documentation, we will also need to make sure we add something to the upgrade_notes.rst. To open up a new section for 3.6, so that we make it clear that there are backwards incompatible changes on deleting action through the CLI and API.

mahesh-orch:
@amanda11 I have added new section for st2 v3.6 and added a note for made change in action delete. Please review. Thanks.

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,15 @@
 Upgrade Notes
 =============
 
+.. _ref-upgrade-notes-v3-6:
+
+|st2| v3.6
+----------
+
+* Action delete API has been modified to delete related files from disk along with
+  de-registering actions from database. For action delete CLI command two arguments
```
amanda11, line 12: @mahesh-orch I'd maybe rephrase slightly on the CLI command. I'd mention that the action delete CLI will now wait for approval to delete on disk before proceeding, unless you use one of the -f or --force commands. As we want to get across the difference to those already using that command. 
If they are using it in a script they need to know that they would need to update to use the one of the new parameters, otherwise their existing scripts would need updating.

docs/source/actions.rst
```diff
@@ -71,6 +71,47 @@ To execute an action manually, you can use ``st2 run <action with parameters>``
    # If you want to add a trace tag to execution when you run it, you can use:
    st2 run core.local cmd=date --trace-tag="simple-date-check-`date +%s`"
 
+Modification in Action Delete API
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+|st2| offers functionality to delete actions/workflows by invoking API. Previously this API
```
amanda11, line 77: Suggest replace "Previously" with "Prior to 3.6 release,"

docs/source/actions.rst
```diff
@@ -71,6 +71,47 @@ To execute an action manually, you can use ``st2 run <action with parameters>``
...
+was only de-registering actions from database. This API has now modified to delete related
```
amanda11, line 78: Suggest, replace "This API has now modified" to "In the 3.6 release, this API was modified"

amanda11 (changes requested):
Looks good, just a few minor comments.

mahesh-orch:
@amanda11 I have done the changes you suggested. Please review.

amanda11 (approved):
LGTM

amanda11:
@mahesh-orch I've not merged this yet. Shall we leave open until the rework on the action to make it default to non-deleting files is done, and the documentation updated.
Or do you want to merge this to match what is there now, and open a new one?

mahesh-orch:
@amanda11, @cognifloyd I have updated the docs according to latest changes in action delete API. Please review. Thanks.

docs/source/actions.rst
```diff
@@ -71,6 +71,52 @@ To execute an action manually, you can use ``st2 run <action with parameters>``
    # If you want to add a trace tag to execution when you run it, you can use:
    st2 run core.local cmd=date --trace-tag="simple-date-check-`date +%s`"
 
+Modification in Action Delete API
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+|st2| offers functionality to delete actions/workflows by invoking API. Prior to 3.6 release
+this API was only de-registering actions from database. In the 3.6 release, with kept backward
```
amanda11, line 78: Perhaps replace "In the 3.6 release, with kept backward compatibility this API was modified to delete related action files from disk as well."
to 
"In the 3.6 release, with backward compatibility retained this API was modified to optionally delete related action files from disk as well."

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,21 @@
 Upgrade Notes
 =============
 
+.. _ref-upgrade-notes-v3-6:
+
+|st2| v3.6
+----------
+
+* With kept backward compatibility, action delete API has been modified.
```
amanda11, line 11: Perhaps rephrase " With kept backward compatibility, action delete API has been modified." to " Retaining backwards compatibility, action delete API has been modified."

amanda11 (changes requested):
Just some minor comments.
However, we could also argue that as this is now backwards compatible it doesn't need to be so detailed in the documents. But I'll leave that to others for their thoughts.

mahesh-orch:
@amanda11 I have made the your suggested changes. Please review. Thanks.

amanda11 (approved):
LGTM / thanks for updating documentation

cognifloyd (approved):

PR #1087, opened 2021-09-01
Pin docutils version 0.16.0

### LINKED ISSUE #1080: Some part of captions are missing on the left side navigation menu.
On the left side navigation menu, Some part of captions (e.g. `Getting Started`) are missing as below.
スクリーンショット 2021-07-24 10 02 09

flunka:
This PR fixes #1080.
Due to some bug in `docutils` 0.17.0 there are missing `caption` class in `<p>` tags.
Pin`docutils` to version 0.16.0 solves the issue.
This is similar issue like here readthedocs/sphinx_rtd_theme#1111

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+docutils==0.16.0
 sphinx>=2.4.4,<3.0
 sphinx-autobuild
 sphinx-sitemap==2.1.0


cognifloyd (approved):
I reviewed the other issues. Using docutils 0.17 is fixed with sphinx 3.5+, but we are using sphinx <3.0 so pinning docutils is a good thing to do.

armab:
@flunka Thanks a lot for the contribution!
First PR and so much highly appreciated fix! :+1: 

Can you also cherry-pick the same change for the https://github.com/StackStorm/st2docs/tree/v3.5 branch?
This way it'll fix it for stable documentation branch: https://docs.stackstorm.com/3.5/

flunka:
I'm glad I could help.
I've created #1088 for v3.5 branch.

PR #1091, opened 2021-10-12
Update install to refer to latest RabbitMQ

amanda11:


docs/source/upgrade_notes.rst
```diff
@@ -8,6 +8,8 @@ Upgrade Notes
 |st2| v3.6
 ----------
 
+* Prior to v3.5 then the installation instructions for all OSes except for CentOS/RHEL 8 were to use the version of RabbitMQ available in the OS distribution. This version is very old, and for 3.6 the installation instructions and simple install have been modified to install the latest version of RabbitMQ. It is not a requirement to upgrade RabbitMQ for installation of 3.6, but for non CentOS/RHEL 8 systems then to keep compatability with a clean installation then RabbitMQ cluster should be upgraded.
```
cognifloyd, line 11:
```suggestion
* Prior to v3.5 the installation instructions for all OSes except for CentOS/RHEL 8
  said to use the version of RabbitMQ available in the OS distribution. This version is
  very old, and for 3.6 the installation instructions--and simple install--have been modified
  to install the latest version of RabbitMQ. It is not a requirement to upgrade RabbitMQ
  for installation of 3.6, but to keep compatibility with a clean installation, the RabbitMQ
  cluster should be upgraded for non CentOS/RHEL 8 systems.
```

cognifloyd (approved):
Looks good. I left a suggestion for one paragraph to increase clarity, fix a couple typos, and fix the rst formatting.

PR #1092, opened 2021-10-19
Rbac key value pair doc

ashwini-orchestral:
Adding documentation for RBAC for key-value pairs for existing and new permission types. Also, permission changes for RBAC KeyValuePair as per logged in user in st2.


m4dcoder (approved):

armab:
Thanks @m4dcoder and @ashwini-orchestral!

PR #1097, opened 2021-11-15
fix indention on advance comparison example YAML 

stevemuskiewicz:
Fixing the indentation on the "advance comparison" YAML block which does not validate.  Updated indentation as a result of [this discussion thread](https://github.com/StackStorm/st2/discussions/5389), which also appears to work when applied in stackstorm.


cognifloyd (approved):
LGTM

PR #1113, opened 2022-03-15
Update documentation for Python 3.8 upgrade on EL8

amanda11:
Update documentation as Python 3.8 will now be used on EL 8 distributions


cognifloyd (approved):
one quick comment. Otherwise, works for me.

Issue #1096, opened 2021-10-29
Migrate docs.stackstorm.com to GH static pages


armab:
Currently, https://docs.stackstorm.com/ are deployed to AWS S3 on every build.
Migrating to native GH static pages unties docs from AWS and would improve the transparency via git, security, and maintenance.
We had a good experience with Github Static pages as a great alternative so far where additional infra would be redundant.

Depends on #1095

cognifloyd:
I think rtfd.io (readthedocs.org) would be ideal for docs.stackstorm.com as they already have the infra to deal with sphinx. Using a CNAME record we can continue using docs.stackstorm.com as well.

armab:
readthedocs.org is really nice, but it's still 3rd party service that will require another new set of credentials and deployment.

With security in mind, what git + GH static pages provide us is native integration and transparency. The entire pipeline doesn't escape Github and everyone can track if the docs.stackstorm.com state is indeed in sync and not modified in the middle or something.

cognifloyd:
> readthedocs.org is really nice, but it's still 3rd party service that will require another new set of credentials and deployment.

Credentials management is easier than I thought it would be with ReadTheDocs. It supports login with github, so we don't need any new passwords to manage this.

As an experiment, I added @st2stanley as an admin of the [st2docs project](https://readthedocs.org/projects/st2docs/). I asked in the readthedocs/community chat about automatically giving a Github Team maintainer rights for a project, and they said:

> for me, direct team access is not a problem. Team members can write the .readthedocs.yml file to update RTD config
> I know is not the same thing...but for 99% of the changes, it works
> this feature is available on our commercial hosting, as GitHub SSO, but we are planning on porting this back to our community platform -- likely starting sometime Q4 maybe?

In the meantime if having @st2stanley as an admin is not enough, we can also add whatever TSC member needs to work on the project. And if we need it to be automated, I can write a GHA workflow that will periodically sync a senior maintainer team (which I don't think we have) from github to RTD.

<hr>

After adding @st2stanley, I tested building docs with the [rtd branch of st2docs](https://github.com/StackStorm/st2docs/tree/rtd). In poking around I noticed that we are using an ancient version of the `sphinx_rtd_theme`: it seems to be somewhere between v0.1.5 and v0.2.0. The theme is at v1.0.0. So, I looked at updating and noticed this warning in the [docs](https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html#via-git-or-download):

[...]

PR #1129, opened 2022-06-07
GitHub action to publish docs to GH Pages

mickmcgrath13:
For https://github.com/StackStorm/st2docs/issues/1096

Things to note:
- This is a first test.  I'll be unable to carry this further; I just wanted to help get it started
- The makefile updates the python version from 3.6 to 3.8 because that's whats provided in the GH Action runner
- The action cannot run until the action config file exists in the base branch of the repo (i.e. `master`)
- Once this is working, we'll still need to update DNS for `docs.stackstorm.com`, but until then, you should be able to go to `stackstorm.github.io/st2docs`.  As a test, check out my forked repo: https://mickmcgrath13.github.io/st2docs/


armab:
@mickmcgrath13 Thanks! :rocket: 

Could you please sign a CLA thing?

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Docs
+
+on:
+  push:
+    branches: [ master ]
```
armab, line 5: We also need versioned branches to be included.

https://github.com/StackStorm/st2docs/blob/6c65852f42ea6a687c1b87e25ad67a320baa1538/.circleci/config.yml#L72-L75

Once merging this PR, the same change should be pushed to all the versioned branches (or at least the last 5-10) to get them generated.
https://github.com/StackStorm/st2docs/branches

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+        # sudo apt install python3-dev
+        sudo apt install libldap2-dev 
+        sudo apt install libsasl2-dev
+        sudo pip3 install virtualenv
+        make docs
+
+        # turn off jekyll so that url routing works properly with underscores
+        touch docs/build/html/.nojekyll
+    - name: Deploy to GitHub Pages
+      uses: crazy-max/ghaction-github-pages@v3
```
armab, line 31: There is another repo that is using `peaceiris/actions-gh-pages@v3` https://github.com/StackStorm/stackstorm.com/blob/3198d86662be419cb5b64197d2223ca225440f20/.github/workflows/gh-pages.yaml#L30-L32

Looking at [`crazy-max/ghaction-github-pages`](https://github.com/crazy-max/ghaction-github-pages) here (300 stars) vs [`peaceiris/actions-gh-pages`](https://github.com/peaceiris/actions-gh-pages) in another repo (3K stars), I think it makes sense to go with a more popular one that we're using already.

.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Build
+      run: |
+        make st2
+        sudo apt-get update
+        # python3-dev is already available
+        # sudo apt install python3-dev
+        sudo apt install libldap2-dev 
+        sudo apt install libsasl2-dev
+        sudo pip3 install virtualenv
+        make docs
```
armab, line 26: Good start, 

Per existing CircleCI config, there's also a need for additional routing logic depending on the branch:
https://github.com/StackStorm/st2docs/blob/6c65852f42ea6a687c1b87e25ad67a320baa1538/.circleci/config.yml#L41-L61

From what I understand the logic looks this way:
* `master` branch build -> `/latest/` st2docs dir
* `v.X.Y.Z` latest stable branch -> `/` st2docs dir
* other `vX.Y.Z` branch build -> `vX.Y` st2docs dir

armab (commented):
@mickmcgrath13 Thanks! :+1: 

I think it makes sense to keep CircleCI just for some time before switching the prod to GH Actions and verifying they're matching 1:1. 

Left some pointers about the branching if you'd like to continue.

.github/workflows/publish-docs.yml
mickmcgrath13: I agree. this has been done.

.github/workflows/publish-docs.yml
mickmcgrath13: full regex filtering does not seem to be supported with GHA: https://github.community/t/using-regex-for-filtering/16427

subset filtering is supported: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

perhaps we could do something like:
```
on:
  push:
    branches:
      - master
      - v[0-9]+.[0-9]+
```

it'd need to be tested, but I think we have another problem with subdirectories (see comment below)

.github/workflows/publish-docs.yml
mickmcgrath13: this might be possible with the `destination_dir: subdir` option

.github/workflows/publish-docs.yml
mickmcgrath13: confirmed the above works.  will resolve this thread when I'm done testing

.github/workflows/publish-docs.yml
mickmcgrath13: so, the `destination_dir: subdir` option works well for `/latest` and `/X.Y` site directories.  so 🎉 

One difference, however, is that if we deploy to the root for the latest GA version like we do for S3, we'll lose all other subdirectories.  For example, if we have `/1.0` and `/latest`, and then we determine that the current version is also the latest GA release, deploying the site to `/` will wipe away `/1.0` and `/latest`.  I've confirmed this with testing.

To get around this, we could implement the action to deploy the latest GA release to a `/ga` subdirectory, and we can use URL routing to point `docs.stackstorm.com` to `docs.stackstorm.com/ga`.  Thoughts?

.github/workflows/publish-docs.yml
mickmcgrath13: https://github.com/StackStorm/st2docs/pull/1129/files#diff-7b112bd01a3c7000d73727a7e66cbfb9cf00a47702c94fda157e23f46e20db17R17

.github/workflows/publish-docs.yml
mickmcgrath13: actually, we have a bigger issue:

When we push to github pages, a bot will run a `pages-build-deployment` pipeline.  If we push to `gh-pages` again while an existing `pages-build-deployment` pipeline is running, github will cancel the previous deployment in favor of the subsequent one:
image

This means that we cannot push to `gh-pages` while there is an ongoing gh pages deployment, and it is non-trivial to "wait" until the gh pages deployment is finished.  I suspect it's possible, though I've not looked into how it might be done (it'd probably involve the github api and some sort of polling logic).

This will also impact us when we want to run a deployment to existing version branches as we'd have to wait until each is finished before triggering the next one..

One thing that might help is that we could utilize the concurrency feature for our workflow.  This will help ensure that our workflows don't run while another of our workflows are running, but we don't seem to have control over the `pages-build-deployment` pipeline.
https://stackoverflow.com/questions/70631168/github-actions-going-through-with-pages-build-deployment

Using `concurrency` might _help_ because it's likely the gh pages will finish before any subsequent builds do, but it's a risk.  It's possible that the gh pages deployment will take some time in which case our st2docs build _might_ finish before and cancel the existing `pages-build-deployment` pipeline.
Concurrency was not working with my testing of multiple jobs in the same workflow, however.. (test here: https://github.com/StackStorm/st2docs/pull/1129/files#diff-7b112bd01a3c7000d73727a7e66cbfb9cf00a47702c94fda157e23f46e20db17R1)

However, it also means that we have to run two builds for the same thing...

One possible workaround is to use routing logic in Route53 or cloudfront to point the latest GA release instead of using a `/ga` subdirectory.  I can think of 2 ways to do this:
1. have some logic in the github action pipeline that modifies entries in route53/cloudfront
2. manually update entries in route53/cloudfront as part of the GA release process

alternatively, it _might_ be possible in the future to publish to GH Pages without the bot, but for now, that's our only option, I think:
https://github.community/t/github-pages-bot-added-to-our-repos-unannounced/218214/13

mickmcgrath13: example pipeline here: https://github.com/mickmcgrath13/st2docs/actions/runs/2531438591

amanda11: i think the issue was that if we are deploying say to version 3.7 of the docs, and then do a deploy to latest whilst its still running the deploy for v3.7. The v3.7 one will be cancelled, and we won't get the update to v3.7.

cognifloyd: But the time `pages-build-deployment` runs, the updated 3.7 sources will already be committed to the gh-pages branch. `pages-build-deployment` redeploys everything in `gh-pages` afaik, so the next `pages-build-deployment` run will include both sets of changes.

mickmcgrath13: from my [previous notes](https://github.com/StackStorm/st2docs/pull/1129#discussion_r901967475):
> One difference, however, is that if we deploy to the root for the latest GA version like we do for S3, we'll lose all other subdirectories. For example, if we have /1.0 and /latest, and then we determine that the current version is also the latest GA release, deploying the site to / will wipe away /1.0 and /latest. I've confirmed this with testing.

One thing that can possibly work around these issues is to always only build gh pages in a pipeline based on the base branch.  the base branch gh-pages pipeline will also `clone` other versions/subdirectories and build those.

then we have pipelines for version-specific branches that just trigger the base gh-pages pipeline to rebuild everything

an all-in-one pipeline at the base branch that builds root and subdirectories would also mitigate special routing requirements (i.e. we just build all the subdirectories we need)

just some thoughts..

PR #1133, opened 2022-07-12
Add notes for sensors and RBAC on upgrade

amanda11:
Update upgrade notes so that mentions sensor_service RBAC assignment is needed if sensors in 3.7 use the key/value store.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -44,8 +44,8 @@ Upgrade Notes
   information.
 
 * As part of extending RBAC support to include protecting access to datastore operations, if
-  you have RBAC enabled and any sensors access the datastore, then the ``sensor_service`` will need to
-  be assigned an RBAC role with the appropriate key_value_pair permissions.  
+  you have RBAC enabled and any sensors access the datastore, then the ``sensor_service`` user will
+  need to be assigned an RBAC role with the appropriate key_value_pair permissions.  
   Further information can be found in the :doc:`RBAC documentation <rbac>`.
 
 .. _ref-upgrade-notes-v3-6:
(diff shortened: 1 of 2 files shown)


amanda11:
Updating the upgrade-notes, as this has come up a couple of times in Slack as a problem on upgrade. Once, we are happy with documentation change then I'll cherry-pick to the 3.7 documentation.

docs/source/upgrade_notes.rst
```diff
@@ -43,6 +43,11 @@ Upgrade Notes
   See :doc:`purging old data documentation <troubleshooting/purging_old_data>` for further
   information.
 
+* As part of extending RBAC support to include protecting access to datastore operations, if
+  you have RBAC enabled and any sensors access the datastore, then the ``sensor_service`` will need to
```
cognifloyd, line 47: Is sensor_service the user?

amanda11: Yes, updated text to make explicit.

jamesdreid: Looks good to me and should address the RBAC access issues with sensors going forward.  Is there a corresponding entry that mentions this or the "sensor_service" user in the RBAC docs for users that have already updated to 3.7 but are just starting to enable RBAC?

amanda11: I couldn't find a mention! I've just updated the PR to add sensor_service to the rbac page - hope that addresses it.

cognifloyd (approved):

docs/source/upgrade_notes.rst
cognifloyd: Thanks. Looks good.

PR #1137, opened 2022-07-18
document output_schema and output_schema masking per StackStorm/st2#5319

cognifloyd:
This adds some more documentation for ``output_schema`` to bring docs up-to-date with StackStorm/st2#5319.

### DIFF
--- a/docs/source/upgrade_notes.rst
+++ b/docs/source/upgrade_notes.rst
@@ -26,8 +26,9 @@ Upgrade Notes
         type: bool
       property2:
         type: str
+        secret: true
 
-  should be updated like this:
+  should be updated to a full JSON Schema like this:
 
   .. code-block:: yaml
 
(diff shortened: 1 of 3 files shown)


docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,43 @@
...
+.. _ref-upgrade-notes-v3-8:
+
+|st2| v3.8
+----------
+
+* For anyone that uses ``output_schema``, which is disabled by default:
+  If you have ``[system].validate_output_schema = True`` in st2.conf AND you have added
+  ``output_schema`` to any of your packs, then you must update your action metadata.
+
+  ``output_schema`` must be a full jsonschema now. If a schema is not well-formed, we ignore it.
```
amanda11, line 15:
```suggestion
  ``output_schema`` must be a full jsonschema now. If a schema is not well-formed, it will be ignored.
```

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,43 @@
...
+  Now, ``output`` can be types other than object such as list, bool, int, etc.
+  This also means that all of an action's output can be masked as a secret.
+
+  To get the same behavior, you'll need to update your output schema.
+  For example, this schema:
```
amanda11, line 20:
```suggestion
  For example, this schema used prior to 3.8.0:
```

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,43 @@
...
+  .. code-block:: yaml
+
+    output_schema:
+      type: object
+      properties:
+        property1:
+          type: bool
+        property2:
+          type: str
+      additionalProperties: false
```
amanda11, line 41: Can we add a comment about when they should install the updated pack, and explain the consequences? e.g. if they install the new version prior to upgrade then the schema will be ignored, and if they update after upgrade then it will be ignored. Ignoring the schema can cause secrets defined in the output schema to be ignored and therefore not masked.

They can use tools such as st2-pack-install to install during the down-time, or ensure that affected workflows are not run until after the upgrade and packs updated.

amanda11 (approved):
LGTM - just some minor suggestions.

rush-skills (approved):

docs/source/upgrade_notes.rst
cognifloyd: K. I added a note.

PR #1142, opened 2022-08-03
Update theme

cognifloyd:
This updates to the latest version of the theme. I believe I've covered all the edge cases, so the sitemap should work, and the versions menu should be backwards compatible (though we need to double check the versions menu once this is merged to be safe).


cognifloyd:
Ok. I've verified that this maintains the versions menu for older versions and I think I've fixed the sitemap. So this should be ready to merge.

I have not removed the old forked theme, we can do that in a follow up once we're satisfied nothing else is missing.

cognifloyd:
This does not as read the docs - it only updates the theme

mamercad (approved):
🚀

cognifloyd:
so, readthedocs is expected to fail - I will follow up after this PR with the bits that allow it to build PRs.

rush-skills (approved):

PR #1143, opened 2022-08-31
WIP: Document using pantsbuild

cognifloyd:
### Background

This is part of introducing [`pants`](https://www.pantsbuild.org/docs), as discussed in the TSC Meetings on [12 July 2022](https://github.com/StackStorm/community/issues/105) and [02 Aug 2022](https://github.com/StackStorm/community/issues/107). Pants has fine-grained per-file caching of results for lint, fmt (like black), test, etc. It also has lockfiles that work well for monorepos that have multiple python packages. With these lockfiles CI should not break when any of our dependencies or our transitive dependencies release new versions, because CI will continue to use the locked version until we explicitly relock with updates.

### Overview of this PR

As we replace the Makefile and other parts of our CI and development infrastructure, the development and contributing guidelines need to be updated. As we integrate pantsbuild in the main st2 repo, we can document changed processes in the `pantsbuild` branch (this PR). Any of the TSC members can push to this branch, so we can collaboratively update the documentation.

### PRs in StackStorm/st2

- StackStorm/st2#5713

### DIFF
--- a/docs/source/development/index.rst
+++ b/docs/source/development/index.rst
@@ -1,6 +1,8 @@
 Development
 ===========
 
+<!-- TODO: update to describe development using pantsbuild -->
+
 This page describes the |st2| development processes and contains general
 guidelines and information on how to contribute to the project.
 


docs/source/development/index.rst
```diff
@@ -1,6 +1,8 @@
 Development
 ===========
 
+<!-- TODO: update to describe development using pantsbuild -->
```
armab, line 4: I think we'll need a small section after `Pre commit hook`

armab:
Just having a couple of commands hinting how to build things with pants for someone residing in the stackstorm repo would be sufficient or perhaps even a more high-level overview of the 2-3 steps that pants does referring to their official docs.

There is no need of a big write-up and explanation and we should stick with the format/size that's already used, like an existing "Code Style Guide".

PR #1147, opened 2022-09-29
Update to use latest erlang (i.e. 25)

amanda11:
Pinning to erlang 24 is causing issues, and erlang 25 is now RabbitMQ's preferred erlang - so use latest instead of pinning to 24.*


nzlosh (approved):
LGTM

PR #1152, opened 2022-11-07
Minor doc changes related to garbage collection

amanda11:
TTL was reduced from 7 days to 1 days a few releases ago
Add upgrade note about new garbage collection option, like what we did for 3.7 new options.


armab (approved):
Nice find!

PR #1156, opened 2023-02-07
Sphinx versions menu fix

cognifloyd:
The versions menu has a duplicate entry for the dev release: "latest" and "3.9dev", and the "3.9dev" link is broken.

Old version menu:
Screen Shot 2023-02-07 at 15 23 46

New version menu:
Screen Shot 2023-02-07 at 15 22 05

I opted to use "3.9dev" to avoid confusion about "latest" meaning stable. I did not change the slug in the url however, only the link text.


rush-skills (approved):
LGTM. I also don't think that using `latest` is a good idea as people will confuse that with stable

PR #1159, opened 2023-04-17
updated

meevelfrancis06:
An open source event-driven platform for  runbook automation is called StackStorm (abbreviation: ST2). It has been compared to SaltStack and Ansible and supports the Infrastructure as Code (IaC) approach to DevOps automation; however, it primarily focuses on doing things or running processes depending on events. Similar to IFTTT or Zapier, StackStorm offers a means of integrating numerous services into cohesive applets or processes that start in response to predefined events or triggers.

While Stackstorm has been used to automate workflows in many other industries, the Arteria project, which offers components to automate analytical and data-management chores at next-generation sequencing core-facilities, is a particularly fascinating application. It uses StackStorm and a micro-services architecture to build an event-driven automation system.

### DIFF
--- a/stacktext.txt
+++ b/stacktext.txt
@@ -1,0 +1,1 @@
+StackStorm documentation


cognifloyd:
I can't find a way to delete this spam, so I will edit the PR description

PR #1161, opened 2023-08-30
Remove note to outdated stuff/links

voxel01:
DC Fabric Automation Suite seems to be not supported anymore, all the links are broken

### DIFF
--- a/docs/source/install/system_requirements.rst
+++ b/docs/source/install/system_requirements.rst
@@ -42,10 +42,6 @@ This is the recommended minimum sizing for testing and deploying |st2|:
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


armab (approved):
Thanks for the contribution!

@voxel01 Could you please sign CLA (instructions above)? So we could merge this.

docs/source/install/system_requirements.rst
```diff
@@ -42,10 +42,6 @@ This is the recommended minimum sizing for testing and deploying |st2|:
 
 .. note::
```
armab, line 43: worth removing the _note_ block as well.
I'm guessing that's why the CI is failing

armab (changes requested):

PR #1170, opened 2024-07-12
Update development docs for Rocky 8.4+ changes

FileMagic:
Update development docs for how to add lsb_release command to rocky 8.4 and up manual build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ DOC_BUILD_DIR := docs/build
 
 BINARIES := bin
 
-PYTHON_VERSION := python3.6
+PYTHON_VERSION := python3.8
 
 # All components are prefixed by st2
 COMPONENTS := $(wildcard st2*)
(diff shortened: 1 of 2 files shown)


FileMagic:
I am going to be going through the whole docs here, so I will make a comment once I have been through all of these docs and have updated stuff.

FileMagic:
I am also going to fix circleci for this...

FileMagic:
> I am also going to fix circleci for this...

I think this was due to the changes in python version.

https://app.circleci.com/pipelines/github/StackStorm/st2docs/699/workflows/7bd58acc-1d9e-4ec2-8a9c-b6749d701f44/jobs/3441?invite=true#step-108-60_40

FileMagic:
I also found out that rocky 9 does not have support for mongo org, so I am not going to make this update to the docs until we switch to a newer version of mongo which  would change these same docs anyways.

PR #1172, opened 2024-09-26
Update sensors.rst to add blurb on get_user_info

FileMagic:
This never seemed to be documented, but I was looking for something similar in the docs: https://docs.stackstorm.com/changelog.html?highlight=get_user_info#id60

### DIFF
--- a/docs/source/sensors.rst
+++ b/docs/source/sensors.rst
@@ -128,6 +128,18 @@ For example:
     self._logger = self.sensor_service.get_logger(name=self.__class__.__name__)
     self._logger.debug('Polling 3rd party system for information')
 
+2. get_user_info(name)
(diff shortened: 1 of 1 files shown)


docs/source/sensors.rst
```diff
@@ -128,6 +128,18 @@ For example:
     self._logger = self.sensor_service.get_logger(name=self.__class__.__name__)
     self._logger.debug('Polling 3rd party system for information')
 
+2. get_user_info(name)
+~~~~~~~~~~~~~~~~~~~
+
+This method allows the sensor instance to retrieve the user can retrieve information
+about the user account which is used to perform datastore operations.
```
cognifloyd, line 135: This is not clear. Could you clean it up?

